PR #12, opened 2024-06-06
Update README.md

vmarks:
correct amble to ample, correct streamling to streamline, and correct some small grammatical choices.


dhh:
I'll take the typos, but the rest were intentional styling choices.

Issue #3, opened 2024-06-04
Add Nord theme


dhh:
This is just gorgeous: https://www.nordtheme.com/

iljo-dp:
I'il do this if you want, working on it right now :D

dhh:
Yes please! We need it across all the 5 apps, and then we need a rights-free wallpaper that isn't too large as well.

PR #14, opened 2024-06-06
Added the nord theme & kanagawa to omakub

iljo-dp:
Haven't tested if the gnome setting works, and obviously i'd need DHH to sign off on the background and the theming. It is all nord be he still needs to like it. 
#3 
Paper is from
https://github.com/linuxdotexe/nordic-wallpapers/blob/master/wallpapers/nord_lake.png
This also looked good imo
https://github.com/linuxdotexe/nordic-wallpapers/blob/master/wallpapers/ubuntu-frost.png


iljo-dp:
if possible i'd like for someone to check all the themes in the respective apps :D

iljo-dp:
Also added the kanagawa theme(trying atleast)
nvim, allacritty and zellij work, but not so sure about gnome, since the theme isn't normally installed, and I can't test gnome at this point in time

refer to 
https://github.com/Fausto-Korpsvart/Kanagawa-GKT-Theme 
is what I'd do, I can't figure it out and test it at this moment in time, sorry

iljo-dp:
I realise the way i do kanagawa isn't the best solution(for gnome), could you give some input on this pull @dhh

dhh:
We don't need any more theming than already done just with colors. So can drop that part of it.

iljo-dp:
now kanagawa should be ready also

dhh:
Thanks a bunch for your help on these! I picked some different backgrounds and tweaked a few settings, but your theme files were great.

PR #19, opened 2024-06-06
Adds unzip dependency required by install/fonts.sh

snth:
This might already be a required by one of the packages on Ubuntu 24.04 but in my Ubuntu 22.04 WSL2 trial unzip wasn't available when it was needed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,7 +3,7 @@ source ~/.local/share/omakub/ascii.sh
 
 # Needed for all installers
 sudo apt update -y
-sudo apt install -y curl git jq
+sudo apt install -y curl git jq unzip
 
 # Ensure computer doesn't go to sleep while installing
 gsettings set org.gnome.desktop.session idle-delay 0


dhh:
I don't want to get into supporting earlier than 24.04 versions of ubuntu, if we can help it. I won't have any means of testing it on a regular basis. Maybe we should code in a hardstop asking for 24.04+ instead.

snth:
I hear you on the point about not taking on the burden of supporting a different Ubuntu version. However if you look at the PR, it's just adding unzip alongside your other tool dependencies and unzip is clearly needed by your fonts.sh script. IDK which package is making unzip available on 24.04 but it appears to be just a happy coincidence that it occurs earlier than the fonts install. IMHO it would make your process more robust by adding it explicitly in install.sh or fonts.sh because if it's already available at that point then it's almost a noop anyway.

dhh:
I see what you mean. Fair enough.

Issue #9, opened 2024-06-06
problem related to zellij when logging in after a fresh install


lvnilesh:
Thank you for omakub. 

This is a fresh install and I note this error

```
error found when loading .profile 

thread main panicked at zellij-client/src/os_input _output.rs:29.43 

could not get terminal attribute: ENOTTY 

as a result the session will not be configured correctly. You should fix the problem as soon as feasible.
```

dhh:
Can you give some more details about your system? I haven't seen that one across everything I've looked at. Maybe @imsnif can help us figure out what's up.

imsnif:
Seems like Zellij is being launched without a controlling terminal. I'm not aware of the particulars of the environment here, but this can for example happen if Zellij is launched from within a background script.

Any context on when we get this error? Every time a terminal is launched, or?

imsnif:
So, a guess for now after looking around the repo a bit: if the conditional generated by [this command](https://github.com/basecamp/omakub/blob/11a66c5cb6cdd2c27c5751dea2d4b64b3ce80b18/defaults/bash/init#L4) is placed in a bash file that gets invoked outside of the Zellij context, this might be causing the error.

Since this setup comes pre-cooked with Alacritty, may I recommend having Alacritty launch Zellij directly rather than having it launched by bash? If my guess is correct this should resolve the issue.

This can be done like so in the Alacritty configuration:
```toml
[shell]
program = "zellij"
# args = ["-l", "welcome"] # optionally, if you'd like to start Zellij with a welcome screen to help manage sessions
```

dhh:
We can definitely switch to that configuration, but there must be something else going on too. I've installed this setup on a bunch of machines, as have others, and not seen that error. But lemme try to switch over anyway!

dhh:
Hmm, on second thoughts, don't love the idea of moving this into alacritty. Since if someone wanted to replace wezterm or kitty or whatever in its place, they'd have to deal with the zellij integration there too by hand. 

Let's see if we can't find what the actual issue is here. Doesn't seem widespread. Can you share more about your setup @lvnilesh? Are you running this in a wm of some kind or?

dhh:
@lvnilesh Try the suggestion to have alacritty start zellij directly instead of via bash.

lvnilesh:
This diff seems to resolve the issue.
```
diff --git a/defaults/alacritty.toml b/defaults/alacritty.toml
index cdb16b7..0654095 100644
[...]

PR #23, opened 2024-06-06
fix for https://github.com/basecamp/omakub/issues/9

lvnilesh:
fixes https://github.com/basecamp/omakub/issues/9


dhh:
Merged the full fix.

PR #31, opened 2024-06-07
Installation: restore 'idle-delay' setting instead of defaulting it

Totto16:
This resets the 'idle-delay' value to the previous value, respecting the uses settings, instead of defaulting to another specific value (300).

This is much more reasonable, since some user might have already changed that and then the installation just changes that, which is not a good thing to do in silence.

It is not that hard to save the original value before the installation and restoring it after

`gsettings get org.gnome.desktop.session idle-delay` returns a variant in the form of `<type> <value>` therefore we need to only store the second part, for which I use cut, since as far as I know, it is always preinstalled (you could use `awk` or other things, but why not go with the easiest tool to complete the job?)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,14 +5,17 @@ source ~/.local/share/omakub/ascii.sh
 sudo apt update -y
 sudo apt install -y curl git jq unzip
 
+# Get the current idle-delay, to restore it later
+IDLE_DELAY=$(gsettings get org.gnome.desktop.session idle-delay | cut -d " " -f 2)
+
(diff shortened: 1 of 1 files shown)


dhh:
The operating premise at the moment is that Omakub is used on a fresh install of Ubuntu, not an existing one. There are a bunch of settings that are just overwritten, like the dock and plenty of other gsettings. I appreciate this, but I think we need to stick with the idea that Omakub is intended for a fresh setup and that it doesn't try to preserve existing user choices. Going to give false assurances to do that in some areas and not others.

Issue #30, opened 2024-06-07
Convert Gnome extension installs from gext to git clone


dhh:
We use 4 Gnome extensions, and we install them using gext. That causes the installer to pause for user input to confirm these installations. That's a pain. We should just clone the extensions straight into the Gnome directory and then turn them on. Then the installer can finish without any intervention from the user.

iljo-dp:
I should just be done, currently testing it on a fresh vm
let me get the vm first

jeremy:
The confirmation dialog is from Gnome itself. Looks like we can bypass confirmation by telling gext to use the filesystem backend instead of dbus: https://github.com/essembeh/gnome-extensions-cli/issues/13#issuecomment-1529160849

dhh:
Oh, even better if we can just stick to gext and do a bypass like that.

dhh:
Sadly the caution on that PR was right. Only 1 out of 4 gnome extensions used worked if installed like that. Will just live with it for now.

PR #34, opened 2024-06-07
fix gnome installs

iljo-dp:
#30 
This should work, but I've only tested it once, so someone else verifying this and testing would be great


dhh:
Tried this through `gext --filesystem`, which does the same thing. Unfortunately 3/4 of the extensions didn't work being installed like that. Think we have to live with the confirmation dialogues for now. If you find a way to do it without that you verify works, feel free to open a PR again.

PR #36, opened 2024-06-07
Drop rbenv and nodenv in favor of mise

dhh:
Just learned about mise! What a great project: https://mise.jdx.dev/

Need to think whether any migration steps are necessary here. The new initializer won't run on existing systems until mise is installed. How best to either do that or communicate that?


dhh:
Migration will happen by prompt to install mise.

navidemad:
What is the advantage compare to asdf ?

brunoprietog:
@navidemad <https://mise.jdx.dev/dev-tools/comparison-to-asdf.html>

Changing from asdf, I love it

PR #29, opened 2024-06-07
Enabling arm64 support for omakub

iljo-dp:
Hey I'm trying to incoporate arm64 support under this script as mentioned in 

Currently NOT every app will work, with the biggest app that doesn't work being chrome
We can go chrome flatpak for example here, but leaving that up to you @dhh
Currently I am NOT finished doing every app and checking every package.
A good rule of thumb, If I haven't changed anything it probability works on arm unless it is farther down the list then docker.sh
Further I haven't checked YET
I haven't also done neovim(even though it is easy) since I was waiting for the other pull request about neovim to be finished before doing neovim stuff in this pull

PR #46, opened 2024-06-08
Exit on error

coffeejunk:
If any of the (sub)commands in `install.sh` fail, the install script will continue to run and effectively hide the fact that something went wrong. Setting `set -e` will make the bash script exit upon an error. 

Context: I was just testing #29 in an `arm64` VM, and the installation of `mise` failed silently (because the current version does not have a code path for `arm64`); however, it appeared as if the installation had worked correctly and I got logged out at the end of it. 
Then, when I tried to log in again, the system would simply hang. Upon further inspection, I realized it was hanging on this prompt:

https://github.com/basecamp/omakub/blob/9c603cd1f72be4c8c1e4b02f2f12ac39359e81a4/defaults/bash/init#L4

Screenshot 2024-06-08 at 11 08 22 AM

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,9 @@
 # Be fancy
 source ~/.local/share/omakub/ascii.sh
 
+# Exit immediately if a command exits with a non-zero status
+set -e
+
 # Needed for all installers
 sudo apt update -y
 sudo apt install -y curl git unzip


iljo-dp:
I think this is a great catch and I think a great implementation and use of the command, btw the issue in #29 should be fixed

josephchigiz:
Yes, this makes for a very important addition.

PR #53, opened 2024-06-09
Update: alacritty default font size from 7 to 10

0ussamaBernou:
font size of 7 is too small for a default value, 10 is more reasonable

### DIFF
--- a/configs/alacritty.toml
+++ b/configs/alacritty.toml
@@ -1,4 +1,4 @@
 import = [ "~/.config/alacritty/theme.toml", "~/.config/alacritty/font.toml", "~/.local/share/omakub/defaults/alacritty.toml" ]
 
 [font]
-size = 7
+size = 10


iljo-dp:
this is dhh's personal opinionated script, and if 7 works for him that is kinda what matters in the end, this HIS script not yours, and imo changing font size is a personal change that is his call and not that of you, me or anyone else here tbf.

j-collier:
If you try it on a framework 13 laptop it's pretty appropriately sized.

gregmolnar:
Omakub is an opinionated setup script and you can adjust this after the installation if you are not happy with the default value. 
@0ussamaBernou would you mind closing the issue?

dhh:
7 is set to look the best on the current Framework 13 screen. But I agree that it's too small for other screens. I'd very much like to come up with a scheme where we hit the right default size by checking PPI or something like that.

We COULD also just set default to 9 (my preference on most screens), but then overwrite it to 7 for that Framework screen.

dhh:
Solved via https://github.com/basecamp/omakub/commit/59eb8fd299062b2361e1d711415f9c9b7824d1b0

0ussamaBernou:
> 7 is set to look the best on the current Framework 13 screen. But I agree that it's too small for other screens. I'd very much like to come up with a scheme where we hit the right default size by checking PPI or something like that.
> 
> We COULD also just set default to 9 (my preference on most screens), but then overwrite it to 7 for that Framework screen.

Yeah I agree, setting font size depending on the PPI would be the optimal solution here.

I'll try to work on it sometime

Also, the solution you ended up with is a good middle ground.

0ussamaBernou:
> 7 is set to look the best on the current Framework 13 screen. But I agree that it's too small for other screens. I'd very much like to come up with a scheme where we hit the right default size by checking PPI or something like that.
> 
> We COULD also just set default to 9 (my preference on most screens), but then overwrite it to 7 for that Framework screen.

Yeah I agree, setting font size depending on the PPI would be the optimal solution here.

I'll try to work on it sometime

Also, the solution you ended up with is a good middle ground.

Issue #45, opened 2024-06-08
Consider adding Shellcheck to the project


anavarre:
I reviewed the PRs and several comments were around code style. One way to solve this once and for all is to have an automated Github Action (as an example) that uses https://github.com/koalaman/shellcheck so all contributions respect the rules imposed by Shellcheck.

I also believe it'd be a great addition to `omakub` itself, but that's likely a different discussion altogether.

PR #70, opened 2024-06-10
Add Shellcheck support via GitHub Actions

anavarre:
Per #45


dhh:
We need master cleaned up first to comply with these rules.

PR #67, opened 2024-06-10
Make neovim background transparent for catppuccin and tokyo-night

hiasinho:
As DHH mentions in #64, neovim's background could be transparent given the 97% opacity in alacritty. I did that in my config already.
This PR addresses transparency for tokyo-night and catppuccin. In light themes, it decreases readability for me personally.

Issue #64, opened 2024-06-10
Make neovim background transparent


dhh:
Our default Alacritty configuration has an ever-so-slight translucency that's really nice without harming readability. It's set to 97% in the config. But neovim isn't translucent in this setup. We need to change something in the theme options, I believe, for it to use the terminal translucency. Let's find and add that.

iljo-dp:
I have a very very non elegant (half) working solution
if this is put in the lazy.lua at the bottom for example it will work

vim.cmd([[
  augroup TransparentBackground
    autocmd!
    autocmd ColorScheme * highlight Normal ctermbg=none guibg=none
    autocmd ColorScheme * highlight NonText ctermbg=none guibg=none
  augroup END
]])
vim.cmd("colorscheme tokyonight")

This works, just tested it, but the biggest problem here is that the theme here is hardcoded, fixing this rn

2KAbhishek:
@dhh can this be closed now? since #71 is merged in 🤔

dhh:
Had to revert #71. Wasn't working on my end. So let's keep open until resolved.

2KAbhishek:
I'll take a look and update the PR

anselal:
require("tokyonight").setup({
      transparent = true,
      styles = {
        sidebars = "transparent",
        floats = "transparent",
      },
    })


replace `tokyonight` is your theme

miniMinn24:
> require("tokyonight").setup({ transparent = true, styles = { sidebars = "transparent", floats = "transparent", }, })
> 
> replace `tokyonight` is your theme

This actually works!

PR #71, opened 2024-06-10
Transparent Neovim background for all themes

2KAbhishek:
As mentioned in https://github.com/basecamp/omakub/issues/64 we want to make neovim background transparent.

https://github.com/basecamp/omakub/pull/67 addresses this for catpuccin and tokyonight.

The solution proposed in this PR applies the same logic to all themes, regardless of their support for a transparent option.

This is achieved by utilizing highlight groups to set the background color to `none` for specific groups.


dhh:
Reverted as it wasn't working properly.

PR #72, opened 2024-06-10
Provide headless installation option for environments such as WSL

snth:
Hi,

I took a stab at separating out the gui apps from the terminal based ones. Given the nice layout of the repo this wasn't too hard.

This seems to work for me on a fresh copy of Ubuntu 24.04 on WSL. Hopefully the main `install.sh` shouldn't be affected as I only renamed scripts.


dhh:
We're not going to pursue a headless installation at this time. There's not enough meat on those bones.

dhh:
That's totally fair, Nick. And I'm thrilled that Omakub can help people discover some of the great tools that we've included here. So feel free to take those discoveries and put them into something other combo that just focuses on this. Omakub needs to remain focused to drive forward, and the focus is on providing a full developer environment based on Linux (Ubuntu) ✌️

PR #77, opened 2024-06-10
Nala: apt replacement

leonardobetti:
This pull request proposes integrating [Nala](https://github.com/volitank/nala) as a package manager wrapper to enhance the user experience for package management tasks.

Benefits:
- improved UX: nala provides a more user-friendly interface compared to traditional apt commands. It removes redundant messages, formats package information more clearly, and utilizes color to visually distinguish between package actions (install, remove, upgrade).
- simplifies the output, making it easier for users to understand the impact of package management commands.

Changes:
- Installs Nala wrapper at the beginning of the script.
- Replaces all apt commands with their corresponding Nala commands.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,7 +6,7 @@ set -e
 
 # Needed for all installers
 sudo apt update -y
-sudo apt install -y curl git unzip
+sudo apt install -y nala curl git unzip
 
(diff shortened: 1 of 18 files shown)


dhh:
Thanks for the suggestion, but I prefer to stay close to the Ubuntu defaults when it comes to this ✌️

PR #79, opened 2024-06-10
Adds transparency for neovim

### LINKED ISSUE #64: Make neovim background transparent

2KAbhishek:
Adds a new file `transparency.lua` to `nvim/plugin/after` that sets the highlight groups regardless of selected colorschme

Also fixes some typos in nbim theme files, causing install loops: lazyvim -> LazyVim

> NOTE: This will not apply to already installed setups as the command to copy the transparency.lua file is part of install script or change the theme for omakub

Fixes #64 

Screenshot 2024-06-11 at 1 05 33 AM

Here's Kanagawa with transparency (trying on a clean lazyvim install and increased transparency)


bin/omakub-theme
```diff
@@ -7,6 +7,12 @@ THEME=$(gum choose "${THEME_NAMES[@]}" --header "Choose your theme" --height 9 |
 # Install theme in Gnome, Terminal, and both default editors
 source $OMAKUB_PATH/themes/gnome/$THEME.sh
 cp $OMAKUB_PATH/themes/alacritty/$THEME.toml ~/.config/alacritty/theme.toml
+
+if [ ! -f ~/.config/nvim/plugin/after/transparency.lua ]; then
+    mkdir -p ~/.config/nvim/plugin/after
+    cp $OMAKUB_PATH/configs/neovim/transparency.lua ~/.config/nvim/plugin/after/transparency.lua
+fi
```
2KAbhishek, line 14: This is here only to address themes not working for existing installs, transparency will be auto applied when changing the theme in that case

install/app-neovim.sh
```diff
@@ -4,8 +4,10 @@ sudo apt install -y neovim
 
 if [ ! -d "$HOME/.config/nvim" ]; then
 	git clone https://github.com/LazyVim/starter ~/.config/nvim
+	mkdir -p ~/.config/nvim/lua/plugin/after
+	cp ~/.local/share/omakub/configs/neovim/transparency.lua ~/.config/nvim/plugin/after/
```
dhh, line 8: Why this new directory?

themes/neovim/kanagawa.lua
```diff
@@ -1,7 +1,7 @@
 return {
 	{ "rebelot/kanagawa.nvim" },
 	{
-		"lazyvim/lazyvim",
+		"LazyVim/LazyVim",
```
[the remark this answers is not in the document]
2KAbhishek: that is causing an installation loop during my tests, all other files are workin correctly other than these two.

the lazy plugin manager is case sensitive IIRC

dhh:
As-is, doesn't work for me. Tried it with Kanagawa theme. No transparency.

install/app-neovim.sh
2KAbhishek: This will allow neovim to load the transparency configs at the end of other configs, making sure all other highlights are loaded first

2KAbhishek:
> As-is, doesn't work for me. Tried it with Kanagawa theme. No transparency.

Did you try this: This will not apply to already installed setups as the command to copy the transparency.lua file is part of install script or change the theme for omakub

2KAbhishek:
Also try doing a theme change to nord or something else once too please

2KAbhishek:
image

Here's Kanagawa with transparency (trying on a clean lazyvim install)

dhh:
Need to merge master and resolve conflict. We should be removing the old transparency options from the themes too.

2KAbhishek:
@dhh done

configs/neovim/transparency.lua
```diff
@@ -0,0 +1,22 @@
...
+vim.api.nvim_set_hl(0, 'NeoTreeNormal', { bg = 'none' })
+vim.api.nvim_set_hl(0, 'NeoTreeNormalNC', { bg = 'none' })
+vim.api.nvim_set_hl(0, 'NeoTreeVertSplit', { bg = 'none' })
+vim.api.nvim_set_hl(0, 'NeoTreeWinSeparator', { bg = 'none' })
+vim.api.nvim_set_hl(0, 'NeoTreeEndOfBuffer', { bg = 'none' })
+
+-- transparent background for nvim-tree
+vim.api.nvim_set_hl(0, 'NvimTreeNormal', { bg = 'none' })
+vim.api.nvim_set_hl(0, 'NvimTreeVertSplit', { bg = 'none' })
+vim.api.nvim_set_hl(0, 'NvimTreeEndOfBuffer', { bg = 'none' })
```
2KAbhishek, line 22: A note: these highlight groups can be extended if/when necessary

You can find the highlight group name with `Telescope highlights` and then set bg to 'none'

This will give you transparency for that hl group 🥂

PR #81, opened 2024-06-10
Add option to select different prompt types

csfh:
Hi @dhh 

In this PR I've added a prompts directory similar to fonts and themes. This will allow a user to change to different kinds of prompts.

I don't suggest Omakub should ship with a ton of prompts, but perhaps a couple of standard ones, the default you have, something that looks like Pure, or my own "Avatar"-style and perhaps even a Powerline.

Additionally, I've also made some changes to that Omakub uses `$OMAKUB_PATH` throughout the code, this allowed me to develop and enable a local version of Omakub next to the one installed in `~/.local/share/omakub`.

Last thoughts, it might make sense to rename `OMAKUB_PATH` to `OMAKUB_HOME` - while this is non essential, it does seem like `_HOME` suffix is more consistent with more libraries in my experiences.

For your convenience I've included a screenshot.

omakub-prompt

Lastly, thanks for creating this. Omakub has renewed my interest in a Linux based setup.

### DIFF
--- a/bin/omakub
+++ b/bin/omakub
@@ -1,11 +1,11 @@
 #!/bin/bash
 
-source $OMAKUB_PATH/ascii.sh
+source "$OMAKUB_PATH/ascii.sh"
 
 if [ $# -eq 0 ]; then
-	SUB=$(gum choose "Help" "Theme" "Font" "Install" "Update" --height 8 --header "" | tr '[:upper:]' '[:lower:]')
+	SUB=$(gum choose "Help" "Theme" "Font" "Prompt" "Install" "Update" --height 8 --header "" | tr '[:upper:]' '[:lower:]')
 else
 	SUB=$1
 fi
 
-[ -n "$SUB" ] && source $OMAKUB_PATH/bin/omakub-$SUB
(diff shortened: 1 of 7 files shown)


csfh:
In the unlikely event you decide to merge this 😉 - please let me know first since I didn't set up my gpg key yet. I'd like to amend the commits with signatures if they're getting merged.

csfh:
Signed both commits.

dhh:
@csfh Can you show what the different prompts look like with some screenshots? Thanks for working on this!

csfh:
Yeah. Let me a maybe two more for a total of four. I'm not a big prompt-aficionado, but I'll try to cover some of the important ground.

csfh:
This is "Avatar" - a prompt inspired by Pure which is for zsh, https://github.com/sindresorhus/pure.

omakub-prompt-avatar

mandarvaze:
@csfh Have you considered https://starship.rs/ for prompts ? It is far superior and cross platform and works will all the shells

csfh:
Good idea. I will look at starship too!

PR #83, opened 2024-06-11
Switch to Docker's apt repo for fresher releases

jeremy:
Ubuntu 24.04 packages Docker 24. Official repo packages Docker 26. Track the latest rather than the distro.


uninstall/docker.sh
```diff
@@ -1,2 +1,3 @@
-sudo apt remove -y docker.io docker-buildx
+sudo apt purge --auto-remove docker.io docker-buildx docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin docker-ce-rootless-extras
+sudo rm -f "${DOCKER_CONFIG:-$HOME/.docker}/cli-plugins/docker-compose"
```
thaJeztah, line 2: This is probably to cleanup from the previous iteration of this script (which installed in `~/.docker/cli-plugins`) correct?

Wondering if it could use a comment and/or instead to provide instructions to manually remove this version if it's present. The idea of the `~/.docker/cli-plugins` directory is to allow users to upgrade to a different version, so they may have installed it not through this script.

install/docker.sh
```diff
@@ -1,15 +1,24 @@
-sudo apt install -y docker.io docker-buildx
+# Remove conflicting Docker packages
+sudo apt purge --auto-remove docker.io docker-buildx docker-doc docker-compose docker-compose-v2 podman-docker containerd runc
```
thaJeztah, line 2: Is it possible users would run this script in situations where they have an existing setup for any of these (for other purposes)?

It's correct that these can conflict with the official packages; just curious if this should either be detected first (are they installed?) and in that case ask the user to confirm, or depend on the package's "Conflicts:" metadata to make them fail if these are present.

install/docker.sh
```diff
@@ -1,15 +1,24 @@
...
+
+# Install Docker engine and standard plugins
+sudo apt install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin docker-ce-rootless-extras
+
+# Give this user privileged Docker access
 sudo usermod -aG docker ${USER}
 
 # Limit log size to avoid running out of disk
-echo '{"log-driver":"json-file","log-opts":{"max-size":"10m","max-file":"5"}}' | sudo tee /etc/docker/daemon.json
+echo '{"log-driver":"json-file","log-opts":{"max-size":"10m","max-file":"5"}}' | sudo tee /etc/docker/daemon.json > /dev/null
```
thaJeztah, line 18: Is the `json-file` logging driver needed for a specific purpose? (i.e. are there tools that try to consume the files it produces directly?).

Generally, [no tools should access those files directly](https://docs.docker.com/config/containers/logging/json-file/?highlight=jsonfile), as they're designed to be internal (and exclusively accessed by ) the daemon, but I know there's some tools that do (so maybe there was a reason for using this);

Screenshot 2024-06-11 at 12 08 26


If that's not the case, then the `local` logging driver may be a good alternative; it can be more performant, and does rotation by default; https://docs.docker.com/config/containers/logging/local/

vvoland: Opened: https://github.com/basecamp/omakub/pull/85

install/docker.sh
dhh, line 2: None of these packages should be installed on a default 24.04, though, right? We haven't hardened anything else against stuff like that. Think we should just base this off the fresh setup, yeah?

thaJeztah: Yes, I don't think any of these would be installed by default. I know we list these in our docs to make users aware that if they have them installed, they may conflict.

But if we assume "fresh install" it's probably better to remove this part. If the packages have the right "conflicts" they should still make the installation fail (which is time for the user to read the error and fix the problem 🥹😅)

jeremy: Cool to target fresh setups only. Should uninstall scripts target current Omakub only, then, or previous versions as well?

dhh: Would just target current version. Uninstall is just a nice-to-have anyway. You can always remove excess stuff by hand.

uninstall/docker.sh
jeremy: Obviated by having uninstall scripts only target the latest incarnation of Omakub installs.

So it'll leave behind stuff you installed with a previous version of Omakub.

install/docker.sh
jeremy: Fine to crash and burn on conflict.

dhh:
We need to have another go at this. I couldn't get it working out of the box with lazydocker connecting correctly on a fresh installation. So reverted it at the last minute before release.

thaJeztah:
Hm.. interesting; do you have more information on what happened? The only thing coming to mind is possibly differences in the systemd unit, with more current versions being configured to wait for [other services / dependencies to be up](https://github.com/moby/moby/blob/v27.0.3/contrib/init/systemd/docker.service#L1-L6)

install/docker.sh
```diff
@@ -1,11 +1,21 @@
...
 
 DOCKER_COMPOSE_VERSION="2.27.0"
 DOCKER_CONFIG=${DOCKER_CONFIG:-$HOME/.docker}
 mkdir -p $DOCKER_CONFIG/cli-plugins
 curl -sSL https://github.com/docker/compose/releases/download/v$DOCKER_COMPOSE_VERSION/docker-compose-linux-x86_64 -o $DOCKER_CONFIG/cli-plugins/docker-compose
-chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
+chmod +x $DOCKER_CONFIG/cli-plugins/docker-compose
```
thaJeztah, line 21: I also just realised that the script is installing compose _twice_ (once from the deb package, and once from the static binaries from GitHub), although that _shouldn't_ be problematic (other than possibly pinning / overriding with an older version).

haberbyte:
I wonder what the problem with docker or lazydocker even is? I just tried it on a fresh Ubuntu installation in a VM and everything works great when installing docker like this 🤷‍♂️

jeremy:
I couldn't get it to fail either. Just works on fresh install, reinstall, and after switching to Ubuntu packages and back. Will reopen a fresh PR.

thaJeztah:
Also had a look at lazydocker to see if anything stood out immediately; it was still using an old version of the docker dependencies, so updated them to more current versions https://github.com/jesseduffield/lazydocker/pull/561#issuecomment-2259830926 (don't think those changes are released yet though).

PR #84, opened 2024-06-11
Disable the line number display in bat command

poshboytl:
By default, batcat displays line numbers, which interferes with multi-line text selection for copying (the line numbers get included in the copied text).

I believe it's better to keep the default behavior the same as cat on line number.

If you think this change isn't reasonable, feel free to just close the request.

### DIFF
--- a/defaults/bash/aliases
+++ b/defaults/bash/aliases
@@ -15,7 +15,7 @@ alias n='nvim'
 alias g='git'
 alias d='docker'
 alias r='rails'
-alias bat='batcat'
+alias bat='batcat --style=plain'
 
 # Git
 alias gcm='git commit -m'


dhh:
If you go plain, what's the main advantage you gain over just using cat?

poshboytl:
@dhh I think we can still enjoy all other benefit like code highlight like below:
Screenshot from 2024-06-11 13-10-49

But from the docs of batcat, if uses plain text, it should not be code highlighted anymore, 

```
-p, --plain
Only show plain style, no decorations. This is an alias for '--style=plain'. When '-p' is used twice ('-pp'), 
it also disables automatic paging (alias for '--style=plain --paging=never').
```

It might be a bug of batcat, anyway I will close this PR first, then try to find a better way to turn off the line number.

PR #85, opened 2024-06-11
docker: use `local` logging driver

vvoland:
Switch docker daemon to use `local` by default for containers as it's more efficient and compresses logs by default.

See more at: https://docs.docker.com/config/containers/logging/configure/


install/docker.sh
```diff
@@ -1,8 +1,8 @@
 sudo apt install -y docker.io docker-buildx
 sudo usermod -aG docker ${USER}
 
-# Limit log size to avoid running out of disk
-echo '{"log-driver":"json-file","log-opts":{"max-size":"10m","max-file":"5"}}' | sudo tee /etc/docker/daemon.json
+# Use local logging driver - it's more efficient and uses compression by default.
+echo '{"log-driver":"local","log-opts":{"max-size":"10m","max-file":"5"}}' | sudo tee /etc/docker/daemon.json
```
jeremy, line 5: Can rely on default log-opts. (The local driver keeps up to 100m per container by default.)

jeremy (approved):

install/docker.sh
thaJeztah: Yes, correct; default [documentation](https://docs.docker.com/config/containers/logging/local/) is Max-size: 20M, and max-files: 5

If that's acceptable, we could remove the `log-opts`, although perhaps it's ok to keep them (it makes the options slightly more discoverable for users if they'd want to tweak them).

thaJeztah (approved):
LGTM

PR #87, opened 2024-06-11
Add php install/uninstall script

### LINKED ISSUE #73: Add default setup for PHP
Omakub should be for developers from a large number of communities, and it should work out of the box for them. What's needed for an out-of-the-box experience for PHP, say, someone developing Laravel?

florentdestremau:
Fixes #73 

- install PHP using apt, with some common extensions (e.g. all 3 main db drivers)
- install composer (package manager)
- ~[RFC: bonus] install the [symfony binary](https://github.com/symfony-cli/symfony-cli)~
- uninstall script following [uninstall codestyle](https://github.com/basecamp/omakub/commit/b42654f70e1c90f02ef001a43680bb2d560ad63b)

According to Laravel docs, the initial setup has all tools included (namely the `artisan` console). I'm not Laravel dev so I can't guarantee it's sufficient.

### DIFF
--- a/install/php.sh
+++ b/install/php.sh
@@ -1,0 +1,18 @@
+# install php with basic extensions
+sudo add-apt-repository -y ppa:ondrej/php
+sudo apt -y install php8.3 php8.3-{curl,apcu,intl,mbstring,opcache,pgsql,mysql,sqlite3,redis,xml,zip}
+
+# install composer, the PHP package manager
+EXPECTED_CHECKSUM="$(php -r 'copy("https://composer.github.io/installer.sig", "php://stdout");')"
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+ACTUAL_CHECKSUM="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"
+
+if [ "$EXPECTED_CHECKSUM" != "$ACTUAL_CHECKSUM" ]
+then
+    >&2 echo 'ERROR: Invalid installer checksum'
+    rm composer-setup.php
+    exit 1
+fi
+
+php composer-setup.php --quiet && sudo mv composer.phar /usr/local/bin/composer
+rm composer-setup.php
--- a/uninstall/php.sh
+++ b/uninstall/php.sh
@@ -1,0 +1,4 @@
+sudo apt -y purge "php8.3*"
+sudo add-apt-repository -y --remove ppa:ondrej/php
+
+sudo rm /usr/local/bin/composer


tonysm:
They're using [Mise](https://mise.jdx.dev/) to install [Node, Ruby, and Go](https://github.com/basecamp/omakub/blob/master/install/mise.sh), can we just use that instead? I've been using ASDF for ages myself, so Mise should just work. It should automatically Composer too. IMO, the PHP setup should be similar to what they did for Go [here](https://github.com/basecamp/omakub/pull/75). So:

- Install PHP via Mise
- Enable the LazyVIM extras for PHP development ([docs](https://www.lazyvim.org/extras/lang/php))
- Install the VSCode extensions for folks that want to use VSCode (I guess only the main ones for Laravel/Symfony development?)

Now, we might need some extra stuff for Laravel, though. The PHP extras are too verbose for my taste and it doesn't seem to cover Blade (Laravel's template engine syntax).

florentdestremau:
I don't see any php docs for Mise. I've been developping for 10 years in PHP and `apt install` was more than enough, I don't see any solution having a _large_ adoption other than Docker-based setups. FrankenPHP is moving up tho, but still early days to make it a default.

I agree I could add the lazyvim extras + vscode. I use phpstorm, in my PHP/Symfony circles this seems to be the default, so I'm not very familiar with other IDE / editors. I have a personal fork with PhpStorm configured but I figure it's maybe too much here

tonysm:
Mise seems to rely on ASDF plugins, so we might be able to `mise use --global php@latest` (or `mise use --global php@8.3` for a specific version) and it should work.

Omakub seems to be either Neovim or VSCode, so other IDEs would have to be installed separately. It's about getting a good experience out of the box with Neovim and VSCode for PHP development too (for the mainstream frameworks, like Laravel and Symfony).

florentdestremau:
The [asdf package for php](https://github.com/asdf-community/asdf-php) doesn't seem very active to me. 

When trying to install with mise I get this error output:

```
configure: error: re2c 1.0.3 is required to generate PHP lexers.
[ERROR] ~/.local/share/mise/plugins/php/bin/install failed
Error: 
   0: ~/.local/share/mise/plugins/php/bin/install exited with non-zero status: exit code 1

Location:
   src/plugins/script_manager.rs:178

Version:
   2024.6.3 linux-x64 (f01cc8d 2024-06-10)

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 2 frames hidden ⋮                               
   3: mise::plugins::script_manager::ScriptManager::run_by_line::hce4bd13bb47001c1
      at <unknown source file>:<unknown line>
   4: <mise::backend::asdf::Asdf as mise::backend::Backend>::install_version_impl::hd8ddf3de6b021ea7
      at <unknown source file>:<unknown line>
   5: mise::backend::Backend::install_version::h31783059b441f187
      at <unknown source file>:<unknown line>
   6: std::sys_common::backtrace::__rust_begin_short_backtrace::h554f9ad47502887c
      at <unknown source file>:<unknown line>
   7: core::ops::function::FnOnce::call_once{{vtable.shim}}::he7b0f19ed095a00d
      at <unknown source file>:<unknown line>
   8: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::h19b9e642d37e7272
      at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/alloc/src/boxed.rs:2020
   9: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once::h97265befc434d3ae
      at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/alloc/src/boxed.rs:2020
  10: std::sys::pal::unix::thread::Thread::new::thread_start::h420dad5cf01a9f35
      at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/sys/pal/unix/thread.rs:108
  11: start_thread<unknown>
      at ./nptl/pthread_create.c:447
  12: __GI___clone3<unknown>
      at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
```

Honnestly this really feels like shoehorning a mediocre solution that isn't quite needed: PHP installs perfectly fine with apt packages  :thinking: . 

As for the editor I totally undestand that vscode and neovim are to be preferred, that's why I'm not including phpstorm here

tonysm:
I just switched from ASDF to Mise here, looks like it works:

```bash
mise use -y -g php@8.3
```

But I had it installed via ASDF, so I guess I had all the dependencies installed already ([here](https://github.com/asdf-community/asdf-php/blob/master/.github/workflows/workflow.yml#L30), as mentioned in the docs).

gcavanunez:
Yeah I'd also +1 opting to use mise over apt given it's what is being used for the other languages ([Ruby, Node, and Go ](https://github.com/basecamp/omakub/blob/master/install/mise.sh)) 

For extras specific to Laravel, aside from adding ` lazyvim.plugins.extras.lang.php` (in `configs/neovim/lazyvim.json`), since Alpine.js is often used, we might be able to get some extras goodies by including the following. 

[From tree-sitter-blade discussion](https://github.com/EmranMR/tree-sitter-blade/discussions/19#discussioncomment-6451292)
`nvim/after/queries/html/injections.scm`
```
;; extends

; AlpineJS attributes
(attribute
  (attribute_name) @_attr
    (#lua-match? @_attr "^x%-%l")
  (quoted_attribute_value
    (attribute_value) @injection.content)
  (#set! injection.language "javascript"))
```
However since I think we're aiming at keeping config low, we may be able to upstreaming to LazyVim, since it looks like the [extras for php](https://github.com/LazyVim/LazyVim/pull/3540) has been added recently.

roelandmoors:
The composer install will stop working if there is a new version.

> WARNING: Please do not redistribute the install code. It will change with every version of the installer. Instead, please link to this page or check [how to install Composer programmatically](https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md).

install/php.sh
```diff
@@ -0,0 +1,13 @@
+# install php with basic extensions
+sudo apt -y install php8.3 php8.3-{curl,apcu,intl,mbstring,opcache,pgsql,mysql,sqlite3,redis,xml,zip}
```
roelandmoors, line 2: - Why install a specific version instead of just 'php'? 
- This will also install apache. Maybe replace `php8.3` with `php-fpm php-cli` ?

florentdestremau: to my knowledge, apache is not included in PHP. PHP8.3 is the current LTS version.

roelandmoors: Without a version it would also work in the next LTS version. Part of apache is a dependency: [PHP](https://packages.ubuntu.com/noble/php) -> [PHP8.3](https://packages.ubuntu.com/noble/php8.3) -> [libapache2-mod-php8.3](https://packages.ubuntu.com/noble/libapache2-mod-php8.3) -> [apache2-bin](https://packages.ubuntu.com/noble/apache2-bin). Not sure if this is a problem for some people. It is just a suggestion I found [here](https://askubuntu.com/questions/1316542/how-to-install-multiple-php-version-without-apache) after trying it on a new Omakub installation

florentdestremau: that's what I personnally use, yes, but specifically because it offers several php versions, I didn't realize it had fewer dependencies too, which is nice. 

As for the LTS, I don't know. Ruby is fixed to 3.3 not latest. PHP 8.4 is going to be released in november, it's quite an easy update to do, LTS are updated every year. It's again a matter of preference, I can't speak for all PHP devs.

roelandmoors: Not sure if it has fewer dependencies but your current PR looks good to me. You have a good point about the fixed versions. I also can't speak for all PHP devs :-)

roelandmoors: Looks like you have to specify a version when using that repository:
image

Maybe you had the better solution from the start. This works:
`sudo apt -y install php8.3 php8.3-{curl,apcu,intl,mbstring,opcache,pgsql,mysql,sqlite3,redis,xml,zip}`

You can remove the version when using the default repository, but I also think adding the ondrej repository is the best solution. 

The Symfony steps are removed, but they work after that and `symfony check:requirements` is successful.

I new Laravel application also works.

osbre:
The installation should also include composer bin directory path so that the globally installed CLI apps work.

```shell
composer global require laravel/installer
```

I.e. this command would make the `laravel` command available in the system provided that composer bin path is properly setup.

roelandmoors:
> The installation should also include composer bin directory path so that the globally installed CLI apps work.
> 
> ```shell
> composer global require laravel/installer
> ```
> 
> I.e. this package would make the `laravel` command available in the system provided that composer bin path is properly setup.

It already does. I tested Laravel and Symfony with sucess.

csfh:
Will definitely review in the morning ET.

osbre:
@roelandmoors I was referring about having this command for `~/.bashrc`:

```shell
export PATH="$PATH:$HOME/.config/composer/vendor/bin"
```

Or for `.zshrc`:

```zsh
export PATH=~/.composer/vendor/bin:$PATH
```

I don't understand how globally installed package binaries would work without having this line. That said, the Symfony CLI you mentioned wouldn't qualify for this test, as it can't be installed via Composer. (unlike Laravel, see the the previous `composer` command)

csfh:
I do believe we need to add the composer bin to PATH.

roelandmoors:
It is already moved to `/usr/local/bin`:
`php composer-setup.php --quiet && sudo mv composer.phar /usr/local/bin/composer`

I tested an earlier version of this PR so maybe I'm wrong, but this seems enough te me?

osbre:
@roelandmoors but that's the composer installation itself, not the vendor bin path (different thing)

csfh:
Yes.

We need the composer executable it self, the .phar file, in `$PATH` which is done with the `mv`.

After that we need to add `export PATH=$(composer global config bin-dir --absolute --quiet):$PATH` to `~/.bashrc` as far as I can tell.

csfh:
I should have included this but I guess this means something like

`echo "export PATH=$(composer global config bin-dir --absolute --quiet):$PATH" >> $HOME/.bashrc" in the installation script, and subsequently, a `sed` or similar to remove it on uninstall.

florentdestremau:
yeah composer is already in the shell path. composer bin, on the other hand, should not be included by default IMO. But that's a matter of preference. I can add that line @csfh if you think it's useful.

csfh:
> yeah composer is already in the shell path. composer bin, on the other hand, should not be included by default IMO. But that's a matter of preference. I can add that line @csfh if you think it's useful.

Hey. I'm not sure why not add it but I would like to hear why you think not be included. It's not up to me at all, but very interesting discussion.

roelandmoors (approved):

florentdestremau:
I would rather leave the $PATH clear of this on install especially when we're in the middle of tens of other installs. I don't really use global composer so didn't think about using it. I agree it could be documented for the php docs here

dhh:
Great work here! Can someone give me a short playbook on the steps I'd have to take with this setup to get "hello world" in Laravel?

csfh:
@dhh 

After you've install PHP using this script, you should be able to run `composer create-project laravel/laravel example-app`. It'll guide you through the rest, including how to eventually start the server. For Laravel, at least locally, there is no need for cgi/apache and so on.

roelandmoors:
I think Laravel is a bit more like Rails, bit if you would like to test Symfony, you can download the cli like this:

`wget https://get.symfony.com/cli/installer -O - | bash`

After that command you get instructions on how to run the Symfony binary or you can just run it like this `~/.symfony5/bin/symfony`

You can then create a complete web app (with importmap/stimulus/turbo as default) like this: 
`symfony new --webapp my_project`

Inside the my_project folder you can start `symfony server:start`

dhh:
Great. I changed this to run inside the new select-dev-env script, but otherwise left all the same. Tested getting a Laravel hello world. Excellent!

Would be nice to get PHP properly supported by mise, though. This just hardcodes PHP 8.3.

florentdestremau:
Thanks for merging :smiley:  For symfony the symfony cli is not required, you can do pretty much the same as for laravel to get a hello world.

```
composer create-project symfony/skeleton:"7.1.*" symfony_app
cd symfony_app
composer require webapp
php -S localhost:8000 -t public
```

As for mise, I discovered it with this project, will try to look at it. I've heard more about nix than for mise for setting up PHP envs.

PR #88, opened 2024-06-11
Add optional Deezer install

florentdestremau:
Spotify is great but some alternatives 🇫🇷 exist :wink:


dhh:
All power to Deezer, but I'm not looking to use the optional path for alternatives. Just additions. Otherwise we'll soon overflow with alternatives. Of course anyone is free, and should be empowered!, to install whatever the hell they want after setting up omakub.

PR #89, opened 2024-06-11
Added lazygit "lg" alias to bash

florentdestremau:
I was happy to discover lazygit, but it's too long to type :stuck_out_tongue:


dhh:
I hear you, but we're already using the l* aliases for ls style shortcuts. Don't love the overlap here. Also, if you're using neovim, you'll have lazygit inside your editor. And otherwise there's always tab completion.

dhh:
Although would consider lzg and lzd for lazygit and lazydocker.

PR #90, opened 2024-06-11
Add Ctrl-N keybinding to open new Alacritty window

r3ap3r2004:
When using a multi-monitor setup, multiple terminal windows can be desirable.  This PR adds an Alacritty keybinding to allow pressing Ctrl-N to open a new Alactritty window.

### DIFF
--- a/defaults/alacritty.toml
+++ b/defaults/alacritty.toml
@@ -12,5 +12,6 @@ opacity = 0.97
 
 [keyboard]
 bindings = [
-{ key = "F11", action = "ToggleFullscreen" }
+{ key = "F11", action = "ToggleFullscreen" },
+{ key = "N", mods = "Control|Shift", action = "CreateNewWindow" }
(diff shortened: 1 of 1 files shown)


dhh:
That unfortunately clashes with the zellij resize mode. But also, we have Alt+shift+2 to start a new terminal (and alt+shift+1 for a new browser). Very handy on different workspaces and monitors!

r3ap3r2004:
Interesting.  On my install resize mode happens with `ctrl+n` and `ctrl+N`, so I didn't see the problem with using the capital `N` for a new window .

dhh:
Ah, I see, you had it on shift N. Yeah, that can definitely work too, but I'd rather keep it consistent. So shift+alt+[number] starts a new instance of both browser and terminal, and it's the same mapping as alt+[number] for switching to them. Totally fine of course to just tailor this after install!

PR #98, opened 2024-06-12
fixed transparent background in nvim

MislavPukljak:
Not all themes ware good configured.
In same themes (tokyonight, catppucciner, nord, rose pine...) not all segments were transparent.
For example:
**Before**
which-key-before
notify-before

**Later**
which-key-later
notify-later


iljo-dp:
lgtm

PR #99, opened 2024-06-12
no more snap for pinta

iljo-dp:
just adding a ppa to remove the snap version of pinta


matsest:
This [ppa](https://launchpad.net/~pinta-maintainers/+archive/ubuntu/pinta-stable) is not maintained anymore:

> Note: this PPA only provides GTK2 builds of Pinta (1.7.x) and is no longer updated for Pinta 2.0 and later. Installing the flatpak or snap package is recommended instead.

It also says in the [doc](https://www.pinta-project.com/howto/installing-pinta) that it requires mono installed.

iljo-dp:
yep fair enough, i did forget to check that, thanks a lot

PR #101, opened 2024-06-12
Add Shellcheck support via GitHub Actions. Fixes all violations.

anavarre:
- Fixed all Shellcheck violations
- Sometimes resorted to using `# shellcheck disable=<shellcheck_code>` immediately above a line to silence violations (per https://www.shellcheck.net/wiki/Directive) since I can't test this code locally (I'm on Mac but supportive of the effort!) and don't want to break things (this should still be tested as I might have missed something).

Background: #70 was reverted because the lift was too much and `master` needed to be cleaned up first.

----

Artifact for disabled violations:

```
❯ ack "shellcheck disable"
install/set-framework-text-scaling.sh
4:# shellcheck disable=SC2063

install/app-signal.sh
4:# shellcheck disable=SC2002

install/set-theme.sh
4:# shellcheck disable=SC1091

install/a-shell.sh
5:# shellcheck disable=SC1091

install/apps-web.sh
3:# shellcheck disable=SC1090

install/docker.sh
7:# shellcheck disable=SC1091

install/select-optional-apps.sh
4:# shellcheck disable=SC2140
8:	# shellcheck disable=SC1090

install/select-dev-env.sh
4:# shellcheck disable=SC2140
28:# shellcheck disable=SC2140

install.sh
15:# shellcheck disable=SC1090
```


install/app-chrome.sh
```diff
@@ -1,5 +1,7 @@
-cd /tmp
+#!/usr/bin/env bash
+
+cd ~/Downloads || exit
```
rmacklin, line 3: Looks like when you resolved conflicts with https://github.com/basecamp/omakub/commit/ec7655e75e6a8806c49b16c624c56ea75ff1eb51, this one wasn't handled correctly: 

```suggestion
cd /tmp || exit
```

anavarre: Nice catch @rmacklin

anavarre:
And here's evidence that Spellcheck fires up fine https://github.com/basecamp/omakub/actions/runs/9482343428/job/26127026453?pr=101

iljo-dp:
I know shellcheck likes this, but imo it isn't really needed, and imo kind of ruins the cleanness of the codebase, because the code will still work
```bash
#!/usr/bin/env bash
```

anavarre:
> I know shellcheck likes this, but imo it isn't really needed,

Explanation here: https://stackoverflow.com/a/25202821

I'd say if we want to start turning off rules that Shellcheck ships with, it's perphaps a sign that we don't want Shellcheck to begin with.

rmacklin:
Disagree that it ruins cleanliness - it's just one comment line in the scripts, and it ensures the scripts will work correctly (by invoking bash) even if the user has switched to a different shell after installation.

install/app-fastfetch.sh
```diff
@@ -1,3 +1,5 @@
-sudo add-apt-repository -y ppa:zhangsongcui3371/fastfetch
+#!/usr/bin/env bash
+
+sudo add-apt-repository ppa:zhangsongcui3371/fastfetch
```
florentdestremau, line 3: shouldn't this line keep the `-y` flag to avoid prompting the user for confirmation?

anavarre: Right, I must have messed this up in my merge conflict. Thanks!

anavarre:
@dhh I keep on rebasing changes (you move fast!), so in order to proceed with integrating the codebase with Shellcheck, if you're still interested with this I think we should merge ASAP.

dhh:
Hmm, looking through the list of changes, I'm not sure I'm seeing enough value here. I like the idea in theory, but the practice of the diff isn't convincing enough imo. Appreciate you investigating this though!

rmacklin:
@dhh Separate from the other changes, would you  accept the addition of a `#!/usr/bin/env bash` shebang to the scripts to ensure they will be invoked with bash even if the user has changed their shell after installation? IMO that part is valuable

PR #102, opened 2024-06-12
Add support for flatpak

hirefrank:
Snap is great but doesn't have everything. Add support for flatpak packages.

### DIFF
--- a/install/app-flatpak.sh
+++ b/install/app-flatpak.sh
@@ -1,0 +1,3 @@
+sudo apt install -y flatpak
+sudo apt install -y gnome-software-plugin-flatpak
+flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo


iljo-dp:
To quote DHH directly when he was asked about flatpak earlier
"No, not interested in flatpak as a general replacement for apt. It's a different philosophy. I prefer apt."

hirefrank:
FWIW, I prefer apt to and I don't think of Flatpak as a replacement anymore than I think of Snap as a replacement. That said not everything is available in apt.

Same is true for AppImages. Unfortunately the ecosystem is fragmented as much as we wish it not to be.

dhh:
I think having flatpak available as an option is actually fine. I just didn't want to convert all our existing installers, or any future installers that can use apt, over to flatpak.

dhh:
Hmm, it's not great that this halts the installation:

image

Is there a way to sidestep that @hirefrank? Otherwise we might just have to make it an optional for now.

PR #103, opened 2024-06-12
Add Gear Lever

hirefrank:
An utility to manage AppImages with ease! Gear lever will organize and manage AppImage files for you, generate desktop entries and app metadata, update apps in-place or keep multiple versions side-by-side.

Dependent on Flatpak support.


dhh:
We're not using any AppImages by default, so I don't think this adds enough to be worth it at this time.

hirefrank:
Understood

PR #104, opened 2024-06-12
enable ufw and turn of iptables for docker

gregmolnar:
Enabling UFW will block all incoming traffic by default and that's a good secure default. @dhh I added a file in the install folder, but this is just turning it on since it is installed by default. Is there a better location for this?
Addtionally, by default, docker exposes the ports on the local network and it can lead to security issues. Disabling the iptables rules docker appends prevents this while still having everything accessible on the host

### DIFF
--- a/install/ufw.sh
+++ b/install/ufw.sh
@@ -1,0 +1,2 @@
+sudo ufw enable
+sudo ufw show verbose
--- a/install/terminal/docker.sh
+++ b/install/terminal/docker.sh
@@ -12,4 +12,4 @@ sudo apt install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin d
 sudo usermod -aG docker ${USER}
 
 # Limit log size to avoid running out of disk
-echo '{"log-driver":"json-file","log-opts":{"max-size":"10m","max-file":"5"}}' | sudo tee /etc/docker/daemon.json
+echo '{"log-driver":"json-file","log-opts":{"max-size":"10m","max-file":"5", "iptables":false}}' | sudo tee /etc/docker/daemon.json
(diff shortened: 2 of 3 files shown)


jeremy (approved):
Closes one can of worms and opens another. There's a decade+ of history to this, unfortunately. Can see why Docker for Desktop (running docker in a VM) is attractive for dev.

I think this is a better starting point than otherwise, though. Surprisingly wide-open ports is heinous.

gregmolnar:
@jeremy Thanks for the review! I resolved the merge conflict.

install/ufw.sh
```diff
@@ -0,0 +1 @@
+sudo ufw enable
```
jeremy, line 1: May be nice to `sudo ufw status verbose` so people can see the firewall status on install.

jeremy (commented):
Consider whether we should set a default policy to deny incoming connections as well, e.g.
```bash
sudo ufw default deny incoming
sudo ufw default allow outgoing
sudo ufw allow OpenSSH

sudo ufw enable
sudo ufw show verbose
```

gregmolnar:
@jeremy ufw by default denies all incoming and allows all outgoing traffic. Since omakub is work workstations, I would also not open SSH. Omakub doesn't even have the SSH server configured as far as I know. 
Showing the status on install is a good idea, I will add that line.

dhh:
Great! Can you add a migration for existing installs as well? So it'll get enabled on the next update.

gregmolnar:
@dhh I added the migration.

jeremy:
Causing dev pains straightaway, of course 😂

Think we'll need something like https://github.com/chaifeng/ufw-docker instead.

/cc @kevinmcconnell @gregmolnar

leepa:
Heads up - this change causes my Docker install not to start because of the iptables config change.

```
unable to configure the Docker daemon with file /etc/docker/daemon.json: json: cannot unmarshal bool into Go struct field Config.log-opt
```

jeremy:
Ha yes, the `iptables` setting is incorrectly embedded in `log-opts`.

jeremy:
Let's back this out entirely and go with https://github.com/chaifeng/ufw-docker instead.

PR #106, opened 2024-06-12
Create a optional install for 1password-cli, required to access secrets via SSH

eddygarcas:


dhh:
Let's just combine this with the 1password installer and install both at the same time.

eddygarcas:
Done, I've just merged both installation scripts into one.

dhh:
If we are setting up the app repo for the cli, is it not possible to use that to install the gui too?

eddygarcas:
> If we are setting up the app repo for the cli, is it not possible to use that to install the gui too?

I'll check this out, I think we can, just didn't want to mess up with the previous configuration. I'll update the PR shortly.

dhh:
image

It's not working as-is.

PR #110, opened 2024-06-12
Added steam as optional app

### LINKED ISSUE #93: Add Steam as an optional install
Linux does gaming! And Steam is the primary platform for distributing those games. Let's get an optional installer setup that does the whole thing. There's a .deb package, but that just kicks off the "real installation". If the optional package is installed, Steam should get added to the dock.

jkjarvis:
Closes #93


dhh:
Very nice. I'd like for us to complete the installation, though. The .deb file only basically sets up the thing to run the full installation. We should trigger the full installation here when you've chosen it.

jkjarvis:
Oh. I did it manually and thought that'll be upto the user. I'll add the steps for full installation.

jkjarvis:
Updated the step with extra packages which were required for installation.
After running the script, the steam icon appears on dock and when a user opens it, we get the login screen

Question: Why does it show the conflict for select-optional-apps, is that file not allowed to be omitted?

dhh:
image

This isn't working right.

jkjarvis:
I didn't get the installer box yesterday , I think first time when I had installed it, it might have kept some cache since uninstalling steam and steam-launcher might not have removed this setting.

I will check this since now have a complete new system with no previous steam installation, thanks

jkjarvis:
Screenshot from 2024-06-14 16-23-53
Screenshot from 2024-06-14 16-24-12

jkjarvis:
Just ran the script and then ran "steam" on the terminal and it started downloading dependencies with manual "enter" , I will check this but I think running "steam" can do the installer part.
After installing, we can kill steam so this credentials window doesn't appear suddenly or should we keep it ?

PR #116, opened 2024-06-13
added command in app-neovim file to disable plug in notifications

MislavPukljak:
This will disable notification after installing lazyvim.
It adds one config to lazy.lua, and didn't make that line to be executed every time because every time will add that sam config.
So I would say the only problem can be that with update command wont be executed, I assume it should be copied and pasted do terminal manually.


dhh:
Seems like we can get away with just #118 👍

PR #121, opened 2024-06-13
Add elixir as a development option

### LINKED ISSUE #92: Add elixir as one of the dev options via mise
I think there's an adsf plugin that can be used to make this happen. Just need someone who actually knows elixir to test it out that everything works as it should.

edgarlatorre:
### Add elixir as dev option

I added elixir as an option then if it's select it install latest erlang and elixir via mise.

https://github.com/basecamp/omakub/assets/52708/b47f0925-4797-496c-91ed-d2e9cbfcc30d

closes #92


dhh:
Awesome. Is this all that's needed? Do you develop with Elixir and can confirm this is sufficient?

edgarlatorre:
> Awesome. Is this all that's needed? Do you develop with Elixir and can confirm this is sufficient?

l think we could install [hex](https://hex.pm/) which is the package manager and maybe [Phoenix](https://www.phoenixframework.org/) for web development?

dhh:
Let's add hex, but stay agnostic on framework.

edgarlatorre:
> Let's add hex, but stay agnostic on framework.

Yes, it makes sense. I'm create a pr adding hex.

PR #122, opened 2024-06-13
added chrome theme change

### LINKED ISSUE #65: Sync theme with Chrome theme extension
Chrome has some very nice theme extensions that really match the look we're going for. It would be great if we could fix or build a matching theme for chrome for our 7 default setups. Here's the Nord theme, for example: https://chromewebstore.google.com/detail/nord-theme/dhlnjfhjjbminbjbegeiijdakdkamjoi?utm_source=ext_app_menu

Chrome stores its preferencs in ~/.config/google-chrome/Default/Preferences. It's a json file. I think we could probably use jq to substitute in the theme we need.

MislavPukljak:
Google Chrome theme can be change when the whole theme is changed, then you need to confirm that you want to change theme in chrome, and chrome restarts.

If somebody thinks that can found better matching colors for chrome, fill free to send config and we can change it easily.
 
Closes #65

### DIFF
--- a/bin/omakub-theme
+++ b/bin/omakub-theme
@@ -12,6 +12,13 @@ if [ -n "$THEME" ] && [ "$THEME" != "<<-back" ]; then
       cp $OMAKUB_PATH/configs/neovim/transparency.lua ~/.config/nvim/plugin/after/transparency.lua
   fi
 
+  # Chrome theme update
+  if gum confirm "Do you want to update Chrome theme? If yes, Chrome will be restarted!"; then
+    pkill -9 chrome
+    source $OMAKUB_PATH/themes/chrome/$THEME.sh
+    google-chrome > /dev/null 2>&1 &
+  fi
+
   cp $OMAKUB_PATH/themes/neovim/$THEME.lua ~/.config/nvim/lua/plugins/theme.lua
   sed -i "s/theme \".*\"/theme \"$THEME\"/g" ~/.config/zellij/config.kdl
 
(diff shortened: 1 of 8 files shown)


dhh:
The background only changes correctly on the first theme change. Subsequent theme changes do not change the background.

dhh:
Also, pkill gives you a "Chrome didn't shut down correctly" error when it restarts.

MislavPukljak:
When you refresh the page or open new tab, background do not change also?

dhh:
So we kill Chrome, and then it starts again. When it starts again, it's not the new background.

MislavPukljak:
Actually do not know what is the problem, I tried multiple time on my laptop and it works.
 
About force kill, that is the only way a menage to change theme, if it is only pkill, or pkill -f, it do not change anything.

dhh:
Think maybe you need to do pkill -f, but then wait until the process is actually dead. Otherwise there's a race condition.

MislavPukljak:
You were right about waiting for process to finish, but that only solved the problem with _"Chrome didn't shut down correctly_", but not with loading new background on restart chrome.

Also find out, when in settings under "On startup" when its selected "_Open the New Tab page"_ then background its loaded instantly. Only when it's selected _"Continue where you left off"_ then it needs to reload current page or open new tab.
I assume that is because that option, chrome opens "old" config, but that is just my guess.

I am trying to find that option to be "force" reloaded, but unsuccessfully for now.

PR #129, opened 2024-06-13
Make alacritty default terminal for CTRL+ALT+T

collinbarnwell:


clneagu:
<code>
sudo update-alternatives --install /usr/bin/x-terminal-emulator x-terminal-emulator /usr/bin/alacritty 50
sudo update-alternatives --config x-terminal-emulator
</code>

collinbarnwell:
> ` sudo update-alternatives --install /usr/bin/x-terminal-emulator x-terminal-emulator /usr/bin/alacritty 50 sudo update-alternatives --config x-terminal-emulator `

Thanks - seems like the first command alone works. I put that in.

csfh (approved):

dhh:
We already have hotkeys for alacritty, so I'm not sure overwriting the default here is necessary. And I've found that it's actually nice to have a way to use the system terminal in case there are any issues with alacritty.

PR #131, opened 2024-06-14
wait till process is done

MislavPukljak:
I apologize for lot of PR's on same thing.

Now next execution is on waiting till chrome is not shutdown properly.

### DIFF
--- a/bin/omakub-theme
+++ b/bin/omakub-theme
@@ -15,7 +15,10 @@ if [ -n "$THEME" ] && [ "$THEME" != "<<-back" ]; then
   # Chrome theme update
   if gum confirm "Do you want to update Chrome theme? If yes, Chrome will be restarted!"; then
     pkill -f chrome
-    wait
+
+    while pgrep -f chrome > /dev/null; do
+        sleep 0.1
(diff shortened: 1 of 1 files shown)


dhh:
That worked!

Next step would be to set the Tokyo theme by default on initial installation.

MislavPukljak:
Have problem with setting default theme in chrome.
Chrome initial settings are set after first time opening the chrome, after that Preferences file is created.

Tried to create manually that file and copy config with theme, but chrome just ignore it and overwrite it.
Also tried to create whole google-chrome directory in .config/, but it just do not work.

I am stuck, so if you will have some advice or idea how to solve this, I can give it a try...

dhh:
We have a similar issue with ulauncher. There we just start ulauncher, then overwrite config. Maybe we can do something similar. Start chrome, then kill it, then overwrite settings.

MislavPukljak:
I tried that, but for config to be created its required to accept some policy and choose between default settings, after that directory Default in google-chrome is created.
That is way I tried to copy configured file and then insert it, but unsuccessfully.

PR #135, opened 2024-06-14
Fix configuring steam after installation

jkjarvis:


jkjarvis:
Screenshot from 2024-06-14 17-55-58
Screenshot from 2024-06-14 17-45-22

jkjarvis:
@dhh I tried the script in virtualbox with a complete new environment and after installing if we run steam in bash “steam -y”, it starts the complete installation process.
There are some prompts in between like press Enter (it was asking for authentication), do we need to automate that ? (not sure how but I can search) Or for now we should expect the user to handle it manually (only 1–2 times it comes) ?

dhh:
Let's try it. Best not to have any user interaction but sometimes there isn't a choice, like with Gnome extensions.

jkjarvis:
Sure, I have removed the custom package installation since when starting steam, it still tries to install them.
Now, the following should happen when running the script:
1. steam.deb download
2. apt update
3. starting steam using script
4. steam download packages
5. Then a UI window opens which updates steam internal components
6. Steam credential window appears.

We can merge this

dhh:
Never did get this to work right, especially through the main install sequence. So punted it to an extra you install after initial setup, and then just rely on Steam's own flow for installing dependencies.

PR #139, opened 2024-06-15
Fix neovim install

roelandmoors:
The lib and share folder should also be copied. The binary is not enough.

Without this PR I get this error when starting neovim after a fresh installation:

image


Nuzair46:
Can be also fixed by `sudo apt install neovim` instead of directly fetching the tar.

roelandmoors:
Yes, but you will have an older version. (0.9 vs 0.10)

csfh (approved):
LGTM! Thanks.

dhh:
Fixed. Thanks for catching this.

PR #142, opened 2024-06-17
Generalizes launching a new instance of dock applications with Shift + Alt + 1/2/..

melopilosyan:
Alacrity and Chrome get special status for the ability to launch a new instance. I believe the Dock platform should provide equal facilities to all residents. 🙂

Besides, the custom shortcuts are tied to the position and the application beyond the Dock's influence. The Alt+# and Shift+Alt+# friendship will be forever broken if I install, say, Brave in second spot.

---

Also changes the behaviour of the Alt + 1/2/3 shortcuts to launch/show and minimize applications with the same keystroke.

### DIFF
--- a/install/desktop/set-dock.sh
+++ b/install/desktop/set-dock.sh
@@ -47,3 +47,6 @@ favorites_list="[${favorites_list%,}]"
 
 # Set the favorite apps
 gsettings set org.gnome.shell favorite-apps "$favorites_list"
+
+# Define the behavior of the app-hotkey binding
+gsettings set org.gnome.shell.extensions.dash-to-dock click-action 'focus-minimize-or-previews'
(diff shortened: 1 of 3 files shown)


dhh:
This is great. Could you rebase, and then we also need to add a migration to remap existing omakub installations.

melopilosyan:
Rebased and added the migration.

Will be happy to implement further suggestions, if any.

dhh:
Hmm. I can't get this working on my setup machine after running the migration. None of the hotkeys work. Not alt+1 or ctrl+alt+1. Trying to run this on a fresh install, but something isn't right.

dhh:
Hmm, yeah, hotkeys don't work on a new install either.

melopilosyan:
My bad.

I'm using a different setup based on Omakub. Thought it would be nice to port this behavior here. But didn't think through the differences. And tested only the migration in VM based on my setup.

Long story short, for the app-hotkey family of key bindings to work, the ubuntu-dock extension has to be enabled. Since that's where these mappings get registered. And Omakub disables ubuntu-dock.

This is what the dock panel looks like on my desktop, centred at the bottom.



It's only visible on the primary display and disappears when application windows overlap its area, including in full-screen mode. In other words, it's there but doesn't get in the way. Plus I can use the app hotkeys.

Let me know if you'd like to try it. I'll open a new PR. This time properly tested in the context of Omakub.

PR #145, opened 2024-06-17
Add newline config to vscode settings

rajraj:
Adding these to the vscode settings,

```
"files.insertFinalNewline": true,
"editor.scrollBeyondLastLine": false
```

### DIFF
--- a/configs/vscode.json
+++ b/configs/vscode.json
@@ -22,5 +22,7 @@
   "explorer.confirmDelete": false,
   "extensions.ignoreRecommendations": true,
   "workbench.colorTheme": "Tokyo Night",
-  "window.titleBarStyle": "custom"
+  "window.titleBarStyle": "custom",
+  "files.insertFinalNewline": true,
+  "editor.scrollBeyondLastLine": false
 }


csfh (approved):

configs/vscode.json
```diff
@@ -22,5 +22,7 @@
   "explorer.confirmDelete": false,
   "extensions.ignoreRecommendations": true,
   "workbench.colorTheme": "Tokyo Night",
-  "window.titleBarStyle": "custom"
+  "window.titleBarStyle": "custom",
+  "files.insertFinalNewline": true,
+  "editor.scrollBeyondLastLine": false
```
florentdestremau, line 27: while I'm all for insertFinalNewLine, scrollBeyongLastLine is not necessary IMO

csfh: Fair enough 👍🏻

dhh:
Hmm, I think this is too specific. We generally don't use a trailing new line.

rmacklin:
@dhh @jeremy I'm a little confused re: "we generally don't use a trailing new line" - isn't ensuring a trailing newline the default behavior when writing a file in vim & neovim? (So my interpretation of this PR was to make vscode consistent in following that unix convention. That way if you create a file in vscode, which is later edited in neovim, there wouldn't suddenly be an unrelated diff adding the trailing newline.)

Issue #32, opened 2024-06-07
bash: hash: hashing disabled


Keremimo:
Hello, first of all loving the setup! Installed it on my laptop for testing for now.

Problem: Every time I open a terminal I get "bash: hash: hashing disabled" error.

The only extra I installed on top of the build is NVM (Node Version Manager) and it causes the same error on boot.

I want to enable hashing to fix this but Googling gave me no solutions.

Thank you for any assistance that you could provide me.

Keremimo:
Update: Band-aided the error by adding "set -h" after "source ~/.local/share/omakub/defaults/bash/rc" in .bashrc

dhh:
Anyone else seeing this? We need set +h in there to allow for app's to have dynamic binstops in ./bin. So can't just yank it.

Keremimo:
Do you think it is because I added NVM even though there was Nodenv? Because I totally didn't notice the Nodenv.

dhh:
Quite possible. I would remove both now and install mise instead. See here: https://github.com/basecamp/omakub/pull/36

jeremy:
This is a warning message, not an error. It's due to [nvm calling `hash -r`](https://github.com/nvm-sh/nvm/blob/1750b8d327d1eb9490ffb92e6201a8ec787b5711/nvm.sh#L3700) which warns when hashing is disabled.

Keremimo:
Indeed. But it actually caused a warning message before the desktop environment loaded, which was quite cryptic. NVM removed and issue is resolved. Thank you!

AmiraliSajadi:
Just fyi, in my case, the warning was triggered by the Miniconda's conda init block in .bashrc that is automatically added during its installation. it seems to invoke hash during startup.

Placing `set -h` _before_ the conda init block fixes the issue for me.

PR #151, opened 2024-06-17
Enable command hashing

AlexZeitler:
`set +h` disables command hashing.

Tools like [nvm](https://github.com/nvm-sh/nvm/blob/master/nvm.sh) are using the `hash` command and this will cause a hint to be displayed everytime the `~/.bashrc` file is executed:

```
bash: hash: hashing disabled
```

Removing `set +h` enables command hashing and avoids the hint.

### DIFF
--- a/defaults/bash/shell
+++ b/defaults/bash/shell
@@ -9,6 +9,5 @@ source /usr/share/bash-completion/bash_completion
 
 # Set complete path
 export PATH="./bin:$HOME/.local/bin:$HOME/.local/share/omakub/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
-set +h
 
 export OMAKUB_PATH="/home/$USER/.local/share/omakub"


csfh:
Does it outright break nvm? Just curious. Omakub does include mise for versioning.

Also you should be able to set `set -h` in your $HOME .bashrc if you install applications that rely on hashall.

My understanding of hashall and it's effects in Bash are somewhat limited. If we change this in Omakub, I'd be very curious to hear from @dhh why it was added in the first place. I suspect it has to do with disabling the bash caching of executable locations.

AlexZeitler:
> Does it outright break nvm?

It doesn't' break anything as far as I can tell.

> If we change this in Omakub, I'd be very curious to hear from @dhh why it was added in the first place. I suspect it has to do with disabling the bash caching of executable locations.

I was asking the Campfire chat if `set +h` has been set by intent but didn't get an answer so far.

To me it's mostly cosmetics because it's somewhat unaesthetic when opening the fancy Zellij window and the first thing you get is this warning :grimacing: 

An alternative suggestion would be a troubleshooting guide where the `set -h` workaround is listed.

kmckelvin:
Omakub installs [mise](https://mise.jdx.dev/) for version management, and that plays nicely with the +h setting. I like having `set +h` as it avoids needing to run `hash -r`.

flamedfury:
`set +h` was mentioned here: https://github.com/basecamp/omakub/issues/32#issuecomment-2155397280

csfh:
Makes sense to me if we close this then? It should be possible to enable it in your own bashrc?

PR #153, opened 2024-06-17
Zellij: copy to clipboard on mouse select

movstox:
If we find this useful, enables copy on mouse select in terminal


MislavPukljak:
This feature is enabled by default. Just select text and it copy it on release.

movstox:
True, thanks!

PR #155, opened 2024-06-18
Add installer and uninstaller for CopyQ Clipboard Manager

kmckelvin:
CopyQ is a clipboard history manager.

Since 1Password is also installed by omakub, the installer will configure CopyQ to ignore passwords copied from 1Password so that information isn't leaked on screen later (e.g. sharing your screen).

The clipboard history can be accessed by default with Super+Shift+V

Based this on a discussion with @AlexZeitler in Campfire

### DIFF
--- a/help/hotkeys.md
+++ b/help/hotkeys.md
@@ -93,3 +93,9 @@ The joy of Linux flows from the keyboard. Training yourself to navigate and comm
 | CapsLock M X | 🎉 | xellebrate |
 | CapsLock M 1 | 💯 | 100%       |
 | CapsLock M T | 🥂 | toast      |
+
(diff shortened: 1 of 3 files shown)


csfh (approved):

dhh:
Went with Pano instead for clipboard management: https://github.com/oae/gnome-shell-pano

kmckelvin:
@dhh given Pano's been removed, is it worth considering this PR again?

dhh:
I just took another look. I can't get over the aesthetics: https://hluk.github.io/CopyQ/. Would rather wait for Pano to become Gnome 46 compatible.

PR #158, opened 2024-06-19
Extension: install weekday start extension

AlexZeitler:
[This extension](https://extensions.gnome.org/extension/1720/weeks-start-on-monday-again/) allows to select the day with which the week starts

image


csfh (approved):

dhh:
Sounds like this will set the default to monday out of the box? Don't think that's going to work. Omakub is going to have more use in the US than not.

AlexZeitler:
The settings can be modified from the command line as described here:

https://github.com/F-i-f/weeks-start-on-monday/issues/13#issuecomment-2210455608

dhh:
Would be nice to have that as a pointer in the documentation. But let's just keep it to that.

PR #159, opened 2024-06-19
Added shift+caps lock as alternative caps lock

### LINKED ISSUE #148: Enable Shift + Caps Lock as Caps Lock as fallback to caps lock being used as compose key
I had originally removed the compose key because I use caps lock regularly (in AZERTY to use for capital letters with accents like `éèçà`).  But I'm trying to learn the emoji shortcuts and realized [we could kinda have it both ways](https://askubuntu.com/a/1488373). I'm ready to make a PR if it's considered a good addition.

image

florentdestremau:
fixes #148 
I also added wink 😉 emoji with `<Multi_key> <m> <i>`

### DIFF
--- a/defaults/xcompose
+++ b/defaults/xcompose
@@ -17,3 +17,4 @@ include "%L"
 <Multi_key> <m> <x> : "🎉" # xellebrate
 <Multi_key> <m> <1> : "💯" # 100%
 <Multi_key> <m> <t> : "🥂" # toast
+<Multi_key> <m> <i> : "😉" # wink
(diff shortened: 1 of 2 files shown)


csfh (approved):

javajosh:
"It's easy enough to change the caps lock mapping in the settings if that doesn't work with your workflow." I attempted to do just that (I prefer caps lock to be mapped to control, and I don't use emojis) through Gnome settings. No error, and it works sometimes, but not others. For example, "ctrl+t" works to create a new tab in a browser using capslock, but if the focus is in a text field, the behavior reverts to compose functionality. 

@dhh I'd be open to a (short) advocacy piece about the utility of emojis and having them at your fingertips like this at all times. I've long used caps lock as control, since it's such a commonly used meta key, but would be open to considering a different layout.

PR #160, opened 2024-06-19
Replace Pinta installation from Snap to Flatpak

osbre:
Via https://flathub.org/apps/com.github.PintaProject.Pinta

> [!NOTE]  
> Needs testing. Hopefully, the priority is right and `flatpak` installs before we use it in other app scripts.


csfh:
Super curious here. What is the difference between snap and flatpak in this case?

jrgirvan:
Why not use pinta's ppa repo?

dhh:
Would take a patch to switch Pinta to ppa repo instead. Not a big fan of snap (or flatpak, really!) when we can avoid it.

PR #167, opened 2024-06-23
feat: docker startup on boot + user docker in current shell

adoreblvnk:
Allow user to use docker in current shell without reloading shell & systemd startup docker on boot as per Docker's post-installation steps.

https://docs.docker.com/engine/install/linux-postinstall/


csfh (commented):
This seems fine to me, but I would avoid the stylistic changes at L5 and L15. Better to open an issue to discuss and a PR to modify the style across the project.

PR #169, opened 2024-06-26
Add toggle-maximized keybind

JxJxxJxJ:
Toggle-maximized keybinding with Super+F


JxJxxJxJ:
Ah mb I mixed the PRs, will do them again

PR #175, opened 2024-06-28
Update hotkeys.md

mandarvaze:
Fix emoji table
Terminal hotkeys are really Zellij hotkeys


dhh:
Fixed on https://manual.omakub.org/1/read/29/hotkeys.

PR #178, opened 2024-06-30
Update README.md

javajosh:
Important to emphasize that this is for Ubuntu 24 NOT 22 everywhere it is mentioned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Omakub
 
-Turn a fresh Ubuntu installation into a fully-configured, beautiful, and modern web development system by running a single command. That's the one-line pitch for Omakub. No need to write bespoke configs for every essential tool just to get started or to be up on all the latest command-line tools. Omakub is an opinionated take on what Linux can be at its best.
+Turn a fresh Ubuntu 24.04 installation into a fully-configured, beautiful, and modern web development system by running a single command. That's the one-line pitch for Omakub. No need to write bespoke configs for every essential tool just to get started or to be up on all the latest command-line tools. Omakub is an opinionated take on what Linux can be at its best.
 
 Watch the introduction video and read more at [omakub.org](https://omakub.org).
 


dhh:
Fresh ubuntu installation implies latest release. Don't think we need to plaster this everywhere.

javajosh:
Please reconsider. The downside to users making an easy mistake is considerable. Omakub installed over Ubuntu 22 on my Framework 13 laptop made the terminal unusable, preventing `cd` and `ls` from working. I've additionally submitted a PR that checks distro and version and aborts if incorrect. You may also prefer the shorter `awk` version:

# Check if the system is Ubuntu and get the version
if ! ubuntu_version=$(lsb_release -rs 2>/dev/null); then
    echo "This script is intended for Ubuntu systems only."
    exit 1
fi

# Minimum required version
min_version="24.04"

# Compare versions using awk
if awk -v ver="$ubuntu_version" -v min="$min_version" '
    BEGIN {
        split(ver, a, ".");
        split(min, b, ".");
        for (i in b) {
            if (a[i] < b[i]) exit 1;
            if (a[i] > b[i]) exit 0;
        }
    }'; then
    echo "Success: Ubuntu version $ubuntu_version meets the minimum requirement of $min_version"
    exit 0
else
    echo "Error: Ubuntu version $ubuntu_version is lower than the required version $min_version"
    exit 1
fi
```
Alternatively, you may want to retain the current behavior as a shibboleth to guard against idiots. However, I don't consider myself (or anyone else in the same boat) an idiot so much as eager to try what I think is a lovely curated set of settings and software. I didn't see Ubuntu 24 as a requirement in the readme, only on the website after-the-fact. I'm sure I won't be the only one to make this mistake.

dhh:
Yeah, I'd take a patch to boot.sh to ensure we're doing this on the right version. That's better than documentation.

Issue #174, opened 2024-06-28
Does Omakub require Ubuntu 24?


javajosh:
I just noticed on the website that "Ubuntu 24 is required. I installed over 22. It doesnt work entirely right. Suggest emphasizing this more clearly.

florentdestremau:
the 2nd line of the website mentions it.

> OMAKUB
> An Omakase Developer Setup for Ubuntu 24.04 by DHH

But it might be useful to check with `lsb_release -r` and display a warning if it's not 24.04

javajosh:
Thanks, yes I just saw that. I think it would be wise for the installer to check the distribution name and version and **_abort_** if the requirements are not met. 

I unwisely installed over Ubuntu 22 and my login shell is failing critically: commands like `ls` aren't working. "Oh My Zsh" and its configuration isn't working either. It looks like `zsh` was uninstalled, and my oh-my-zsh configs deleted. 

I'm not entirely sure what to do, not knowing Ubuntu very well. I'm not entirely sure what Omakub did, or how to undo it. I'm not sure if I can get out of this situation by upgrading Ubuntu 22 to 24 (which seems not to have any official docs (!) but there is an [unofficial guide](https://www.cyberciti.biz/faq/how-to-upgrade-from-ubuntu-22-04-lts-to-ubuntu-24-04-lts/). Another alternative is to back out of omakub entirely, but I don't see uninstall directions, and it's not clear if your install script is reversible. Advice on this point is appreciated.

Note that, with some exceptions (for example Alacritty is missing entirely), the desktop is working so I can get into home directory files to make edits with GUI tools. Additionally, it looks like omakub uninstalled `zsh` entirely so at first I was unable to run a subshell. I was able to reinstall `zsh` with `sudo apt install zsh` and get a subshell with working `ls`. So that's something, I guess.

javajosh:
I was able to get out of trouble by upgrading Ubuntu 22 to 24 using [these instructions](https://www.cyberciti.biz/faq/how-to-upgrade-from-ubuntu-22-04-lts-to-ubuntu-24-04-lts/), and then reinstalling Omakub. Note that all Ubuntu 22 users will be able to upgrade more easily on August 15, or so I hear. But I needed a working machine now.

[...]

PR #181, opened 2024-07-03
Update install.sh to abort installation on non ubuntu and non ubuntu …

javajosh:
Omakub is verified to be incompatible with Ubuntu 22, and in fact has disastrous side-effects. See [this issue](https://github.com/basecamp/omakub/issues/174) for details.

The primary install script now checks distribution name and version and aborts when either is incompatible. This change is compatible with future versions of Ubuntu.


dhh:
Let's separate this out into an external script that can exit if the version isn't matched. Would like to keep install.sh clean. See the new choices.sh for an example.

javajosh:
Master moved under me, and complicated the commit history slightly. If this is a problem I can create a new PR. 

This is also a chance to get feedback on the code itself. I suspect you might want to change the name of the script, perhaps `check-distro.sh`. I also had the thought that this might be good to do in `boot.sh` prior to `clone` however you expressed your desire for cleanliness, which means calling it (unconditionally) from `install.sh` after the `clone`.

javajosh:
The file change list is not clean on this PR. I'm going to close this PR and make a new one later tonight when people have stopped committing to master.

PR #187, opened 2024-07-04
Generalize `boot.sh` to support running any ref

rmacklin:
This will enable running not only the `master` and `stable` branches [^1], but also unmerged commits, for easier testing, e.g.

```sh
export OMAKUB_REF=2f1544548d8709f61e9e6aac5e222b000770898b
wget -qO- "https://raw.githubusercontent.com/basecamp/omakub/${OMAKUB_REF}/boot.sh" | bash
```

The hope is this would provide a good way to have folks test things like #29 before they've been merged.

Additionally, together with:
- https://github.com/basecamp/omakub-site/pull/4

it removes the need for a separate `boot-dev.sh` script (which was nearly identical to `boot.sh`).

[^1]: Note: `stable` and `master` will still be runnable via the existing `wget -qO- https://omakub.org/install | bash` and `wget -qO- https://omakub.org/install-dev | bash` commands, respectively. No workflow changes needed for those two common cases!


rmacklin:
This way, an arbitrary ref can be tested like:
```sh
OMAKUB_REF=7963327db7895002e5a5e8d0bc3537334b311608
eval "$(wget -qO- https://raw.githubusercontent.com/basecamp/omakub/7963327db7895002e5a5e8d0bc3537334b311608/boot.sh)"
```

rmacklin:
@dhh Curious to get your thoughts on this. I just updated the PR description to clarify that this would _not_ take away the common case of running
`wget -qO- https://omakub.org/install | bash` or
`wget -qO- https://omakub.org/install-dev | bash`.

The goal is simply to support a similar mechanism for testing other changes that aren't merged to `master` or `stable` (like the work-in-progress changes to add ARM support, as one example where I see this being useful). And a little side-benefit is it obviates the need to maintain a separate, nearly identical `boot-dev.sh` script.

That said, let me know if you have any objections to this, so I can either try to address them or abort and close the PRs.

dhh:
How would you be able to test something that isn't on master through this? Wouldn't those PRs be on a separate cloned repo URL?

rmacklin:
I should clarify that in order to test a PR like https://github.com/basecamp/omakub/pull/29 or https://github.com/basecamp/omakub/pull/227 using this method, the PR would have to be rebased or have `master` merged into it.

dhh:
This is breaking the update script. Since you're no longer directly on the stable branch. Can you have a look at how we might fix this? And also how we might fix it for folks who've installed this in a way where update now won't run?

rmacklin:
I've opened #240 which I think should address this going forward. Unfortunately, I think folks who did a first-run install in v1.1.2 would have to manually run `cd $OMAKUB_PATH && git checkout stable` before updating omakub...

PR #189, opened 2024-07-05
add option to install dotnet

zBrianW:
This PR adds the option for installing dotnet


dhh:
Doing dotnet development on Linux seems like a very niche thing. Is this the domain you work in yourself?

zBrianW:
Thanks for taking the time to review. Yes, I have used in Production, admittedly I would be in the minority, but I'd like to think if given more exposure there could be more adoption (kinda 🐔🥚), but I can appreciate where you're coming from.

dhh:
I think this is too niche for the moment, especially given the fact that we're not really saving you from much work anyway. Thanks for exploring it regardless!

PR #199, opened 2024-07-09
Factor out version check code into a separate script. Call the script…

javajosh:
This is a redo of my previous [PR](https://github.com/basecamp/omakub/pull/181). The git history is clean now, and the code is improved: distro name is properly checked, and the awk version comparison code is simplified.

The main purpose of this PR is to help others avoid [the footgun of installing omakub over Ubuntu 22](https://github.com/basecamp/omakub/issues/174), which is the current LTS release.


check-version.sh
```diff
@@ -0,0 +1,30 @@
...
+            return 1
+        fi
+    else
+        echo "Error: Unable to determine OS. /etc/os-release file not found." >&2
+        return 1
+    fi
+}
+
+if check_ubuntu_version; then
+    echo "Script is running on Ubuntu 24.04 or higher. Continuing execution..."
```
dhh, line 25: Let's just stay quiet if operating system is correct. That's the default assumption. So only output anything if we're failing.

javajosh: I agree. Done.

check-version.sh
```diff
@@ -21,9 +21,7 @@ check_ubuntu_version() {
     fi
 }
 
-if check_ubuntu_version; then
-    echo "Script is running on Ubuntu 24.04 or higher. Continuing execution..."
-else
+if ! check_ubuntu_version; then
     echo "Script execution failed due to system requirements not being met." >&2
     exit 1
 fi
```
javajosh, line 27: Remain silent on version check success.

Issue #201, opened 2024-07-09
Allow one-line theme switch 


florentdestremau:
You can't run `omakub theme "Tokyo Night"` and make it switch directly to this theme. This would help #200 as I already use a hack to override gum's feature in a subscript.

```
( gum() { if [[ $1 == "choose" ]]; then echo "$DESIRED_THEME"; else command gum "$@"; fi; }; source $OMAKUB_PATH/bin/omakub-theme )
```

dhh:
You can see how it's done in sub/font.sh. I think that's a good pattern for this.

dhh:
Feel free to open a PR with this if you choose to explore.

PR #206, opened 2024-07-10
feature: add auto light/dark theme switch

### LINKED ISSUE #200: Light/dark mode switching
I made a local test to try and have 2 themes, one dark and one light:
ezgif com-optimize

Here's my quick script (based on omakub 1.0 for theme switching). Would you be intereseted in incorporating it ? Don't know how this would work, I guess a LIGHT_THEME and DARK_THEME could be saved somewhere ? Also, the monitoring of gsettings could probably be cleaner but it Just Works for now :smile: 

```
#!/bin/bash

# Function to execute when dark mode is enabled
dark_mode_on() {
	echo "Dark mode enabled, setting \"Tokyo Night\" theme"
	export DESIRED_THEME="tokyo-night"
	set_theme
}

# Function to execute when light mode is enabled
light_mode_on() {
	echo "Light mode enabled, setting \"Red Pine\" theme"
	export DESIRED_THEME="rose-pine"
	set_theme
}

set_theme() {
	# Create a temporary gum function that returns the desired theme
	function gum() {
		if [[ $1 == "choose" ]]; then
			echo "$DESIRED_THEME"
		else
			command gum "$@"
		fi
	}

	# Run the script with the overridden gum function
	( gum() { if [[ $1 == "choose" ]]; then echo "$DESIRED_THEME"; else command gum "$@"; fi; }; source $OMAKUB_PATH/bin/omakub-theme )
}


check_mode() {
    local mode=$(gsettings get org.gnome.desktop.interface color-scheme)
    if [[ $mode == *"prefer-dark"* ]]; then
        dark_mode_on
    else
        light_mode_on
    fi
}

# Initial check
check_mode

# Monitor changes in the settings
while true; do
    current_mode=$(gsettings get org.gnome.desktop.interface color-scheme)
    if [ "$current_mode" != "$previous_mode" ]; then
        previous_mode=$current_mode
        check_mode
    fi
    sleep 1
done
```

florentdestremau:
This adds a script that watches the gsettings to detect light or dark mode, and creates a systemd service that runs it in the background. When a switch is detected, it triggers the corresponding theme change.

Fixes #200 

Possible issues (non blocking):

- you have to edit the file to configure the two themes. For now I defaulted to Tokyo Night / Rose Pine, open to feedback on where to store other settings.
- I only tested on wayland, and there is some part of specific in the script (I debugged some issues because of this)
- There is some error output in the service logs because the `source $OMAKUB_PATH/bin/omakub-sub/menu.sh` at the end of the `theme.sh` script crashes with my workaround for applying a theme without prompting "gum choose" . It's not an issue at all but might raise some eyebrows. I made #201 to do this in a cleaner way.

How to test this PR:

- checkout this branch on your local  omakub instance, or copy the files directly
- run `install/desktop/optional/app-auto-dark-mode.sh`
- try switching dark/light mode
- confirm the theme switch
- revert with the uninstall script and delete the files if needed to cleanup

### DIFF
--- a/bin/dark-mode-switch.sh
+++ b/bin/dark-mode-switch.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# Function to execute when dark mode is enabled
+dark_mode_on() {
+	echo "Dark mode enabled, setting \"Tokyo Night\" theme"
+	export DESIRED_THEME="tokyo-night"
+	set_theme
+}
+
+# Function to execute when light mode is enabled
+light_mode_on() {
+	echo "Light mode enabled, setting \"Red Pine\" theme"
+	export DESIRED_THEME="rose-pine"
+	set_theme
+}
+
+set_theme() {
+	# Create a temporary gum function that returns the desired theme
+	function gum() {
+		if [[ $1 == "choose" ]]; then
+			echo "$DESIRED_THEME"
+		else
+			command gum "$@"
+		fi
+	}
+
+	# Run the script with the overridden gum function
+	( gum() { if [[ $1 == "choose" ]]; then echo "$DESIRED_THEME"; else command gum "$@"; fi; }; bash $OMAKUB_PATH/bin/omakub-sub/theme.sh )
+}
+
+
+check_mode() {
+    local mode=$(gsettings get org.gnome.desktop.interface color-scheme)
+    if [[ $mode == *"prefer-dark"* ]]; then
+        dark_mode_on
+    else
+        light_mode_on
+    fi
+}
+
+# Monitor changes in the settings
+while true; do
+    current_mode=$(gsettings get org.gnome.desktop.interface color-scheme)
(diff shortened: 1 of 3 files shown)


florentdestremau:
@dhh a short feedback would help me out here: is this is something you would
1.  consider merging as it is
2. consider merging with some tweaks (the configurable theme choices would be the first one I'd work on anyway)
3. not consider this at all

I'm not on the campfire chat for the project so I don't know how much traction this feature idea has 😀, so I'm waiting for feedback as to dig in further or not. It's been working on my laptop for about a week, no issues noted.

dhh:
I appreciate the work you've put into this, but I think it's too much for Omakub. If you end up creating a separate project for this, though, we could consider linking to it. Thanks for exploring it in either case!

florentdestremau:
okay ! I'll try to package this into a simple script to launch after install. This could be a cool section in the README here, "available extensions" 😀

PR #211, opened 2024-07-12
Fix eza installation

albski:
Catched an error of `Unable to locate package eza`.

Fixed by implementing code from https://github.com/eza-community/eza/blob/main/INSTALL.md#debian-and-ubuntu


dhh:
Sounds like maybe you're trying to run omakub on something other than the official 24.04 release? It's included in 24.04: https://www.ubuntuupdates.org/package/core/noble/universe/base/eza

Issue #195, opened 2024-07-07
Expose colorscheme shell variable


JxJxxJxJ:
I used to have other apps' (lunarvim) configuration files change according omakub's changes through the `theme.sh` script. 

Now after trying to update I noticed many of these files changed and I saw how unpractical my approach was.

I think it'd be useful if omakub exposed a colorscheme variable so I could make my own scripts while not changing omakub's original files, making the update process more seamless.

dhh:
Yeah, I've been thinking of having a ~/.config/omakub/settings.json or similar that would retain state, like current theme. Feel free to investigate and open a PR if you arrive at something pleasant!

PR #212, opened 2024-07-13
Add OMAKUB_THEME shell variable!

JxJxxJxJ:
This is very useful for syncing other apps with the omakub's theme script. #195


JxJxxJxJ:
I think it makes sense to do the same with the fonts and font size. Let me know if I should make another PR

dhh:
How do you use this to sync? Since it'll only be set when changing the theme? I had thought that maybe we should try to store state, like ~/.config/omakub/settings.json. But concerned that it's too easy for things to drift.

JxJxxJxJ:
After sourcing the theme.sh script the variable gets stored in the session, then in the same shell one can set up a different script that reads that variable and syncs other apps. I guess a settings.json makes things easier though, I'll try to take a look into that, maybe using jq + bash is enough for that?

PR #215, opened 2024-07-14
Define env EDITOR var before starting Zellij

gothicx:
The command "vim" is not working.

Documentation:

"By default, Zellij looks for an editor defined in the EDITOR or VISUAL environment variables (in this order). Make sure one is set (eg. export EDITOR=/usr/bin/vim) before Zellij starts."


dhh:
Can't set the editor in the defaults like that, since it wouldn't be changeable. Needs to be in the generated files, if at all.

PR #218, opened 2024-07-15
Update app-zellij.sh not to crash if `~/.config/zellij/themes` exists

jean-francois-labbe:
If the `.config/zellij/themes` exists omakub install fails.

This PR adds a check on the presence of `~/.config/zellij/themes` before attempting to create the folder

### DIFF
--- a/install/terminal/app-zellij.sh
+++ b/install/terminal/app-zellij.sh
@@ -5,6 +5,8 @@ sudo install zellij /usr/local/bin
 rm zellij.tar.gz zellij
 cd -
 
-mkdir -p ~/.config/zellij/themes
+if [ ! -d "$HOME/.config/zellij/themes" ]; then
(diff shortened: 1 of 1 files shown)


rmacklin:
Do you have a nonstandard `mkdir`? `mkdir -p` is specifically designed to [not error if the directory already exists](https://man7.org/linux/man-pages/man1/mkdir.1.html#:~:text=no%20error%20if%20existing)

jean-francois-labbe:
I installed omakub on 11/07/2024.
I tried to rerun it few days ago and faced this issue.

It seems that the `~/.config/zellij/themes` is a simlink.
image

The error message:
Screenshot from 2024-07-15 23-20-35

jean-francois-labbe:
I ran `wget -qO- https://omakub.org/install` from https://omakub.org

Now I'm seeing that the manual has an updating section, I should have tried that

PR #219, opened 2024-07-16
Fix window title and class in .desktop files

mustafabulut34:
Updated .desktop files to correctly specify window title and class for better user experience.


dhh:
What does this change exactly? Can you post a screenshot or similar?

mustafabulut34:
For example,
This is what activity looks like when super+tab + alt+tab,
- without --class and --title
image
image

-  with --class and without --title
image
image

- with --class and --title
image
image


The same goes for whatsapp. Without --class, it looks like a chrome window.

PR #224, opened 2024-07-19
Add Flutter support

CillianMyles:
## Motivation

Canonical mentioned that [Flutter is their default choice](https://ubuntu.com/blog/flutter-and-ubuntu-so-far) for new desktop and mobile apps, having re-designed and [re-built the Ubuntu installer with Flutter](https://canonical.com/blog/how-we-designed-the-new-ubuntu-desktop-installer). 

They list Flutter alongside Typescript and React in their Web Frontend Engineer [job requirements](https://canonical.com/careers/5150422).

As Flutter is technology that can build and deploy to the web, but also given it's ability to deploy to native platforms, and the fact that is very much a first class citizen in Ubuntu and backed by Canonical, I think Flutter deserves a home in Omakub.

### PR includes

- [x] install script
- [x] uninstall script
- [x] vscode support
- [x] neovim support

### Target Platforms
Out of the box this installation gives you the ability to target web and desktop linux:

image

### Editing & LSPs
Both the VS Code and Neovim setups appear to be working (but I am by no means a Vim user, so if anyone would like to validate this I would appreciate it).

### Links
- https://code.visualstudio.com/docs/editor/command-line
- https://github.com/akinsho/flutter-tools.nvim

### DIFF
--- a/install/terminal/select-dev-language.sh
+++ b/install/terminal/select-dev-language.sh
@@ -40,6 +40,9 @@ if [[ -n "$languages" ]]; then
 		Java)
 			mise use --global java@latest
 			;;
+		Flutter)
+			source ~/.local/share/omakub/install/terminal/flutter.sh
+			;;
 		esac
 	done
 fi
--- a/install/terminal/select-dev-language.sh
+++ b/install/terminal/select-dev-language.sh
@@ -2,7 +2,7 @@
 if [[ -v OMAKUB_FIRST_RUN_LANGUAGES ]]; then
 	languages=$OMAKUB_FIRST_RUN_LANGUAGES
 else
-	AVAILABLE_LANGUAGES=("Ruby on Rails" "Node.js" "Go" "PHP" "Python" "Elixir" "Rust" "Java")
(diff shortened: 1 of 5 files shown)


install/terminal/select-dev-language.sh
```diff
@@ -40,6 +40,9 @@ if [[ -n "$languages" ]]; then
 		Java)
 			mise use --global java@latest
 			;;
+		Flutter)
+			source ~/.local/share/omakub/install/terminal/flutter.sh
+			;;
```
CillianMyles, line 45: I am presuming this will load and execute the script. Will try to figure out a way to test it locally. :thinking:

dhh:
Hmm, appreciate this, but I don't want to get into installing specific frameworks, whether they be for web or desktop, as a general approach. There's just too many. Let's keep things at the language level. Thanks for looking into it, though!

CillianMyles:
@dhh I can understand the sentiment of not wanting to get involved in the framework wars.

I'm actually glad I did the PR anyways because I learned a lot about combining bash scripts, and finally turned my own personal config into a single line install 🎉

https://github.com/CillianMyles/dev-config

PR #227, opened 2024-07-22
Raspberry pi compatibility

IbrahimTanyalcin:
I wanted to try Omakub in my raspberry pi5, however ran into premature termination of the script in multiple places. These modifications should not change anything for amd64 but allow installation of most of the apps in raspberry. In the end I was able to have all the good bits working except these:
- chrome (I think chrome is not possible to be installed in arm64 yet anyway. )
- signal
- spoftify
- zoom
- dropbox

So I finally was able to try Omakub like in the introductin video, and it is really cool! Window tiles and shortcuts are great!

Here are the changes:

- added a `utils.sh` to deal with vendor imcompatibilities when naming architechtures amd64 vs x86_64, aarch64 vs arm64 etc. `dpkg --print-architechture` is not enough to deal with all the incosistencies so there is an extendable `arch_config` variable for future apps.
- in raspberry, `XDG_CURRENT_DESKTOP` is not only `GNOME` but `ubuntu:GNOME`, so made some changes to deal with possible variations without breaking current behavior
- since chrome is not possible in arm64, I added Brave to the optional apps that gum asks during installation.
- removed all hardcoded mentions of `amd64` and `x86_64` and replaced them with output of `get_arch`, the output defaults to lscpu's values but can be customized based on app by adding keys to `arch_config` array in `utils.sh`
- sourcing installers are moved to if conditionals which do not trigger `set -e` (I undestand the intention of putting set -e but when something goes wrong you are left with a broken install anyway). So if one of the apps fail, the parent script will continue unless there is something else in the parent script that triggers `set -e`, that behavior is left unchanged. The failed apps are reported in a formated way inside the root folder called `omakub_report.log`. 
- add reporters in `utils.sh` called `handle_omakub_source_error` and `handle_omakub_source_success` which are used by `print_omakub_report | tee ~/.local/share/omakub/omakub_report.log`, this will list all the successful and failed ones. Currently in rasberry pi 5, I only had 5 failed which some of them do not have arm64 releases. 

These changes should not introduce any breaking changes or modifications for `amd64` but will allow people to install and give Omakub a try in their arm64 devices and track what failed or not down the line. 

*** Please test this in your `amd64` machines and verify I did not accidentally introduced breaking changes. ***


rmacklin:
Hey @IbrahimTanyalcin can you rebase off the latest `master`, which includes some changes to make testing unmerged PRs easier.

IbrahimTanyalcin:
@rmacklin Thanks for letting me know, I've rebased, there were no merge conflicts.

install/terminal/mise.sh
```diff
@@ -1,7 +1,11 @@
+if [[ -z $UTILS_SOURCED ]]; then
+    source ~/.local/share/omakub/utils.sh
+fi
+ARCH=$(get_arch)
```
kjeldahl, line 4:
```suggestion
ARCH=$(get_arch "mise")
```
And an entry in utils.sh as well.

IbrahimTanyalcin: `get_arch` at here:  https://github.com/IbrahimTanyalcin/omakub/blob/raspi5-compat/utils.sh#L80-L94 will default to `lscpu`'s output if no entries are found here: https://github.com/IbrahimTanyalcin/omakub/blob/raspi5-compat/utils.sh#L37-L54

I think you want it to be more explicit, in that case I can add these indentity entries to the assoc array here: https://github.com/IbrahimTanyalcin/omakub/blob/raspi5-compat/utils.sh#L38 like: 

```shell
["mise.aarch64"]="aarch64"
["mise.x86_64"]="x86_64"
```

I only added values to that array for the programs that deviate from lscpu's standard arch notation. But if you believe being explicit is better I will do that. What do you think?

dhh:
Closing in favor of pursuing this goal in #259, which is much smaller in scope.

PR #233, opened 2024-07-25
feat: always install latest gum version

NikolaiKotikov:


dhh:
Let's use the same style as in lazydocker/git.

NikolaiKotikov:
@dhh fixed it, take a look please

dhh:
Thanks for getting this updated. But on second thought, I'd actually rather stick with a known version of gum, so we don't have a floating dependency that might break the entire installer if the API changes.

PR #241, opened 2024-08-04
Replace `Vitals` with `TopHat`

Kasui92:
Since `Vitals` installation has been disabled for Intel processors, I think `TopHat` can be a great replacement as a monitoring extension. 

I searched for `TopHat` issues to see if anyone reported any malfunctions with Intel processors, similar to the one highlighted, but found nothing.

I made `TopHat` as similar as possible to the default `Vitals` configuration.


Kasui92:
Close this PR to use a feature branch instead of fork's master. 

Sorry!

PR #244, opened 2024-08-08
Refactor check-version script

soni17:
Refactoring check-version script to avoid nested conditionals and to make easier to read.

### DIFF
--- a/install/check-version.sh
+++ b/install/check-version.sh
@@ -1,27 +1,21 @@
 #!/bin/bash
 
-# Function to check if running on Ubuntu 24.04 or higher
-check_ubuntu_version() {
-  if [ -f /etc/os-release ]; then
-    . /etc/os-release
-    if [ "$ID" = "ubuntu" ]; then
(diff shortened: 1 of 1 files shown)


dhh:
Nice! Can you rebase?

soni17:
rebased

install/check-version.sh
```diff
@@ -1,27 +1,21 @@
...
+if [ ! -f /etc/os-release ]; then
+    echo "$(tput setaf 1)Error: Unable to determine OS. /etc/os-release file not found."
+    echo "Installation stopped."
+    exit 1
+fi
+
+. /etc/os-release
+
+# returns 1 if true and returns 0 if false
+correct_version=$(echo "$VERSION_ID >= 24.04" | bc)
```
dhh, line 12: I think we can actually just inline this check in the conditional below.

soni17: moved it to the conditional

PR #245, opened 2024-08-08
Replace `Vitals` with `TopHat` 

Kasui92:
Since `Vitals` installation has been disabled for Intel processors, I think `TopHat` can be a great replacement as a monitoring extension. 

I searched for `TopHat` issues to see if anyone reported any malfunctions with Intel processors, similar to the one highlighted, but found nothing.

I made `TopHat` as similar as possible to the default `Vitals` configuration.

### DIFF
--- a/bin/omakub-sub/theme.sh
+++ b/bin/omakub-sub/theme.sh
@@ -2,17 +2,18 @@ THEME_NAMES=("Tokyo Night" "Catppuccin" "Nord" "Everforest" "Gruvbox" "Kanagawa"
 THEME=$(gum choose "${THEME_NAMES[@]}" "<< Back" --header "Choose your theme" --height 10 | tr '[:upper:]' '[:lower:]' | sed 's/ /-/g')
 
 if [ -n "$THEME" ] && [ "$THEME" != "<<-back" ]; then
-	cp $OMAKUB_PATH/themes/$THEME/alacritty.toml ~/.config/alacritty/theme.toml
-	cp $OMAKUB_PATH/themes/$THEME/zellij.kdl ~/.config/zellij/themes/$THEME.kdl
-	sed -i "s/theme \".*\"/theme \"$THEME\"/g" ~/.config/zellij/config.kdl
-	cp $OMAKUB_PATH/themes/$THEME/neovim.lua ~/.config/nvim/lua/plugins/theme.lua
+  cp $OMAKUB_PATH/themes/$THEME/alacritty.toml ~/.config/alacritty/theme.toml
+  cp $OMAKUB_PATH/themes/$THEME/zellij.kdl ~/.config/zellij/themes/$THEME.kdl
+  sed -i "s/theme \".*\"/theme \"$THEME\"/g" ~/.config/zellij/config.kdl
+  cp $OMAKUB_PATH/themes/$THEME/neovim.lua ~/.config/nvim/lua/plugins/theme.lua
 
-	source $OMAKUB_PATH/themes/$THEME/gnome.sh
-	source $OMAKUB_PATH/themes/$THEME/vscode.sh
+  source $OMAKUB_PATH/themes/$THEME/gnome.sh
+  source $OMAKUB_PATH/themes/$THEME/tophat.sh
+  source $OMAKUB_PATH/themes/$THEME/vscode.sh
 
-	# Forgo setting the Chrome theme until we might find a less disruptive way of doing it.
-	# Having to quit Chrome, and all Chrome-based apps, is too much of an inposition.
-	# source $OMAKUB_PATH/themes/$THEME/chrome.sh
+  # Forgo setting the Chrome theme until we might find a less disruptive way of doing it.
+  # Having to quit Chrome, and all Chrome-based apps, is too much of an inposition.
+  # source $OMAKUB_PATH/themes/$THEME/chrome.sh
 fi
 
 source $OMAKUB_PATH/bin/omakub-sub/menu.sh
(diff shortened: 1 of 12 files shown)


dhh:
I like that idea. Can you post a screenshot of what it looks like when it's just showing the default download speed?

Kasui92:
> I like that idea. Can you post a screenshot of what it looks like when it's just showing the default download speed?


@dhh  
Sure!

This is the freshly installed default version of the extension.
immagine

This is the version with only the network monitor, without icons and other resources monitor. Unfortunately it is not possible to show ONLY the download speed. 
immagine
immagine

italopenaforte:
I have installed using the script in my old laptop and is working flawless.

LorenzoMugnai:
Hi, I love this idea to use TopHat. 

In my own setup, I’ve already implemented something similar where the TopHat metrics colour adjusts dynamically based on the active Gnome theme.

Here’s how I’ve integrated it in the`omakub/themes/set-gnome-theme.sh` :

```
# Set TopHat color based on the current theme
color="#1dacd6" 
case $$OMAKUB_THEME_COLOR in
    "'purple'")
        color="#924d8b"
        ;;
    "'red'")
        color="#e92020"
        ;;
    "'blue'")
        color="#208fe9"
        ;;
    "'sage'"|"'bark'")
        color="#78ab50"
        ;;
    "'magenta'")
        color="#e920a3"
        ;;
    *)
        ;;
esac
gsettings set org.gnome.shell.extensions.tophat meter-fg-color "$color"
```

This is how it now looks like

Screenshot from 2024-08-14 14-30-45

dhh:
Can you add a migration script so folks will get this installed and vitals removed on the next upgrade?

italopenaforte:
Nice touch @LorenzoMugnai, was not aware that can be done.

LorenzoMugnai:
> Nice touch @LorenzoMugnai, was not aware that can be done.

Thank you! I'm glad you found it helpful. This project has been a great learning experience for me as well.

For my setup, I primarily use numeric mode, so I've disabled the charts using the following commands:

```
gsettings set org.gnome.shell.extensions.tophat cpu-display numeric
gsettings set org.gnome.shell.extensions.tophat mem-display numeric
```

However, it's good to know that if I decide to enable the charts, they integrate nicely with the system. Additionally, I've disabled the animations with:

```
gsettings set org.gnome.shell.extensions.tophat show-animations false
```

Now you have all my settings in case you'd like to integrate them into your setup. 

I hope it helps!

Kasui92:
@LorenzoMugnai 
Yours is a very interesting addition, especially to maintain coherence with the chosen theme. 

I took this opportunity to also add extensions theming to Omakub, maybe in the future it will be useful to have it ready, and then I added `TopHat` metrics color management to this system.

@dhh 
Obviously with your approval!

Kasui92:
@dhh 
Regarding the migration script, do I add a file in `/migrations` or leave a script here? 
If so, what is the naming convention for files in `/migrations`?

Kasui92:
@dhh 
> Can you add a migration script so folks will get this installed and vitals removed on the next upgrade?

Here it is!

```bin/bash
# Uninstall Vitals
if [ -n "$(gnome-extensions list | grep Vitals@CoreCoding.com)" ]; then
    gnome-extensions uninstall Vitals@CoreCoding.com
fi

# Install TopHat
gext install tophat@fflewddur.github.io

sudo cp ~/.local/share/gnome-shell/extensions/tophat@fflewddur.github.io/schemas/org.gnome.shell.extensions.tophat.gschema.xml /usr/share/glib-2.0/schemas/
sudo glib-compile-schemas /usr/share/glib-2.0/schemas/

# Configure TopHat
gsettings set org.gnome.shell.extensions.tophat show-icons false
gsettings set org.gnome.shell.extensions.tophat show-cpu false
gsettings set org.gnome.shell.extensions.tophat show-disk false
gsettings set org.gnome.shell.extensions.tophat show-mem false
gsettings set org.gnome.shell.extensions.tophat network-usage-unit bits

# Set TopHat metrics color to match the theme
THEME_NAMES=("Tokyo Night" "Catppuccin" "Nord" "Everforest" "Gruvbox" "Kanagawa" "Rose Pine")
THEME=$(gum choose "${THEME_NAMES[@]}" "Default" --header "Choose your theme" --height 10 | tr '[:upper:]' '[:lower:]' | sed 's/ /-/g')

if [ -n "$THEME" ] && [ "$THEME" != "default" ]; then
  source $OMAKUB_PATH/themes/$THEME/extensions.sh
fi
```

gregmolnar:
@Kasui92 Can you save that snippet into the migrations folder like the existing ones?

Kasui92:
@gregmolnar Done!

themes/set-gnome-extensions-theme.sh
```diff
@@ -0,0 +1 @@
+gsettings set org.gnome.shell.extensions.tophat meter-fg-color $EXTENSIONS_TOPHAT_COLOR
```
dhh, line 1: I would make this specific. I see the point that what if in the future we might have other extensions that could need styling, but that's future programming. Today we just have tophat. So should just frame all of this around tophat styling 👍

dhh: So that would mean themes/*/tophat.sh and since it's only a single gsettings call, I would just do it inline in each of those files. Don't need the set-tophat-theme.sh helper.

Kasui92: @dhh Done! 👍

dhh:
We're missing a dependency. I just ran the migration and got this:

image

Kasui92:
@dhh 
My bad! 
I did not include these libraries installation in the migration file: 

```
gir1.2-gtop-2.0 
gir1.2-clutter-1.0
```

I included them in the installation process, but I didn't pay attention to them during the migration. 

This should be added before the TopHat installation process, in the file: 

```
# Install Tophat libraries
sudo apt install -y gir1.2-gtop-2.0 gir1.2-clutter-1.0
```

I have the commit on this branch, but the PR is closed. Shall I open a new one?

dhh:
Ah yes. Please do open another PR 👍

dhh:
(Worked after installing the dependency 👍)

PR #248, opened 2024-08-13
Add move-to-workspace hotkeys

Kasui92:
I think these hotkeys can be useful to have to keep alive the philosophy of using the keyboard as much as possible, instead of the mouse. 

To move open applications between workspaces and have everything organized, these hotkeys may be what the user needs.

The added shortcuts are: 

```
Shift+Super+1 - Move application to Workspace 1
Shift+Super+2 - Move application to Workspace 2
Shift+Super+3 - Move application to Workspace 3
Shift+Super+4 - Move application to Workspace 4
Shift+Super+5 - Move application to Workspace 5
Shift+Super+6 - Move application to Workspace 6
```


italopenaforte:
I took a while to discovery those hotkey, will be nice to add them into the manual. 

How we can add them to the manual? @dhh

dhh:
It's in the manual: https://manual.omakub.org/1/read/29/hotkeys 👍. These hotkeys are already mapped via Space Bar extension: https://github.com/basecamp/omakub/blob/master/install/desktop/set-gnome-extensions.sh#L63

PR #252, opened 2024-08-22
Fix missing `TopHat` libraries in migration

Kasui92:
This commit fixes the TopHat libraries missing error after running the migration script.

[Reference](https://github.com/basecamp/omakub/pull/245#issuecomment-2304633085)


Kasui92:
@dhh 
Add the libraries installation and I also added a logout request to avoid that during the migration it doesn't load all the TopHat dependencies correctly in the current user session.

PR #254, opened 2024-08-25
Use flatpak for Pinta installation

Kasui92:
Resolve a `FIXME` left in the Pinta installation, replacing `snap` with `flatpak`.


dhh:
I was actually hoping we could install Pinta directly from a deb. Not through either flatpak or snap. Not convinced that going flatpak over snap actually offers much if any advantage for this otherwise.

Kasui92:
@dhh 
Seems legit! I ll try to switch into .deb .

PR #256, opened 2024-08-31
Add aliases to bash for vi/vim

gothicx:
Those are the aliases most people use, including myself.


dhh:
Omakub comes with a whole suite of hotkeys and aliases that are new to folks. I think that's OK. And also OK for folks to add their own to .bashrc after installation. But I'd prefer to have one omakub default for everything.

PR #258, opened 2024-09-02
restore original idle and lock settings instead of defaults

adrianomitre:
At the end of the installation, restore the original idle-delay and lock-enabled Gnome settings instead of default values.


dhh:
Omakub is only intended for installation on a fresh Ubuntu setup. All sorts of things may go wrong or be overwritten if you run it on a system you've already been tailoring. So I don't think this makes sense in that context. Thanks for investigating, though!

PR #261, opened 2024-09-05
Docker + ufw integration

jeremy:
We disabled iptables integration in #104. Time to reintroduce it with ufw managing what Docker exposes.

/cc @gregmolnar


gregmolnar:
Thanks for fixing this @jeremy . I saw the issues but didn't have the time to look into it.

PR #263, opened 2024-09-07
Update app-localsend.sh

1gn0r4nd:
use flatpak instead of getting deb file


1gn0r4nd (commented):
Added sudo

dhh:
Prefer to avoid both flatpak and snap whenever we can. The deb file is preferable.

PR #267, opened 2024-09-07
feat: add atuin as history tool

gothicx:


dhh:
That's pretty neat, but I think the ctrl-r history search we get from fzf is enough for out-of-the-box Omakub for now.

Issue #266, opened 2024-09-07
when running ufw.sh in a fresh installation it gives an error


jeremies:
the error is the following:

ERROR: UFW is disabled or you are not root user, or mismatched iptables legacy/nf_tables, current iptables v1.8.10 (nf_tables)

when running the line "sudo ufw-docker install" of file ufw.sh

i think you should enable ufw before ufw-docker install. Like this:

![image](https://github.com/user-attachments/assets/938034fe-d47c-4118-837e-8d9064eb5e7b)

dhh:
cc @jeremy

dhh:
Hmm, even with that fix, I still can't get it to work. After doing an install, "sudo ufw status" says that it's off. Will have to yank until we can get it proper sorted.

dhh:
Yanked ufw for now. Will bring back when better tested.

Xamnalim:
Hi, i run into this issue while trying to install ufw-docker on Pop OS 24.04 (distro also based on ubuntu).
The issue for me was that my locales were set to Polish (LANGUAGE=pl_PL_UTF-8).
The ufw-docker script runs the following check:
```sh
if ! ufw status 2>/dev/null | grep -Fq "Status: active" ; then
    die "UFW is disabled or you are not root user, or mismatched iptables legacy/nf_tables, current $(iptables --version)"
fi
```
It checks if the `ufw status` returns `Status: active`, but with non English LANGUAGE env var set, the output was different than expected, even tho the ufw firewall was active.

I don't know if the root cause of the opened by @jeremies was similar, but I thought I would let you guys know.

Possible fix: add `export LANGUAGE=en_US.UTF-8` (or similar English locale) to the begining of the script.

PR #268, opened 2024-09-08
Remove ufw until we can sort it out proper

dhh:
Not working as intended at the moment. As it were, ufw wasn't getting turned on after new installs. See https://github.com/basecamp/omakub/issues/266.


dhh:
cc @jeremy

PR #156, opened 2024-06-18
Dev Env Tweaks

tonysm:
### Changed

- Moves the Golang NeoVim extras plugin config to the dev env script (closer to choosing the lang)
- Enables the PHP extras plugin in NeoVim ([docs](https://www.lazyvim.org/extras/lang/php)) when choosing PHP and define the LSP to intelephense instead of the default (phpactor) - intelephense is much less verbose, IMO, it's the one I prefer using myself and also the one I use in VSCode (there's a comparison [here](https://yarnaudov.com/phpactor-vs-intelephense.html)), I made a patch to LazyVim to support it (https://github.com/LazyVim/LazyVim/pull/3691)
- Enables the Typescript extras plugin ([docs](https://www.lazyvim.org/extras/lang/typescript)), which actually works for all JavaScript environments

PR #269, opened 2024-09-09
Make the LazyVIM extras more modular based on the installed dev language

tonysm:
### Changelog

- Removes the base `lazyvim.json`
- Enable the LazyVIM extras plugin when the language is installed

---

I started this work at https://github.com/basecamp/omakub/pull/156 but it was mixed with changing PHP's default LSP on LazyVIM. I'll send a separate PR for that.


install/terminal/select-dev-language.sh
```diff
@@ -15,16 +25,22 @@ if [[ -n "$languages" ]]; then
...
+			enable_lazyvim_extras "lazyvim.plugins.extras.lang.go"
 			;;
 		PHP)
 			sudo add-apt-repository -y ppa:ondrej/php
 			sudo apt -y install php8.3 php8.3-{curl,apcu,intl,mbstring,opcache,pgsql,mysql,sqlite3,redis,xml,zip}
 			php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
 			php composer-setup.php --quiet && sudo mv composer.phar /usr/local/bin/composer
 			rm composer-setup.php
+
+			enable_lazyvim_extras "lazyvim.plugins.extras.lang.php" "lazyvim.plugins.extras.lang.typescript"
```
florentdestremau, line 43: not sure the typescript should be included here ? :thinking:

tonysm: I added it because I usually write JavaScript too in my PHP apps

florentdestremau: but then the node js option would take care of this, right ?

tonysm: It might, but these days I'm not using Node (using Importmaps)

tonysm: I've removed the TS extras plugin from the PHP installation (I can always enable it myself)

install/terminal/select-dev-language.sh
```diff
@@ -6,6 +6,16 @@ else
...
 fi
 
+enable_lazyvim_extras() {
+	local temp_file=$(mktemp /tmp/omakub.XXXXX)
+	local extras=("$@")
+	local jq_extras=$(printf '"%s",' "${extras[@]}")
+	jq_extras="[${jq_extras%,}]"
+
+	jq --argjson extras "$jq_extras" '.extras |= (. + $extras | unique)' ~/.config/nvim/lazyvim.json >"$temp_file" &&
+	mv "$temp_file" ~/.config/nvim/lazyvim.json
```
dhh, line 16: Rather than the temp file dance, I would just touch ~/.config/nvim/lazyvim.json first, then assume it's there.

tonysm: We can't read and write to the same file in a pipeline like that, something to do with it using the same file descriptor. I tried here, and the config file ended up empty.

However, I found some tricks to achieve this behavior, as described [here](https://backreference.org/2011/01/29/in-place-editing-of-files/). I've updated the code to use one of the tricks described there and it looks like it works.

tonysm: (happy to revert it to use a tmp file, if you don't want to use this trick, btw)

dhh:
If we're adding these LazyVIM extras on install, I think we need to think about having a matching uninstaller.

tonysm:
@dhh I've made the extra plugins optional and also added an uninstall script for the dev languages.

install/terminal/select-dev-language.sh
```diff
@@ -1,45 +1,64 @@
 # Install default programming languages
 if [[ -v OMAKUB_FIRST_RUN_LANGUAGES ]]; then
-  languages=$OMAKUB_FIRST_RUN_LANGUAGES
```
dhh, line 3: Let's stick with 2-spaces-no-tabs.

dhh:
Excellent work! Let's just fix the indention and we're good to go 👌

tonysm:
@dhh I'm done with the edits. Let me know if you have any other concerns.

dhh:
@tonysm If you feel like exploring upgrading the overall uninstaller so it's nicer and more like the installer, please do pursue!

dhh:
Ugh. Turns out that the lazvim.json file doesn't exist until you've run neovim for the first time. Which you won't have done during the first install. So need to initialize this file manually for it to work in first-run mode.

tonysm:
Oh, let me send another one that touches it before modifying it then.

tonysm:
hmm, I didn't find a way to ensure that file is created. I tried just touching the config, but it doesn't work. Also tried creating an empty one and one with just `{"extras": []}` but it also didn't work (when LazyVIM initializes, it merges the existing config with the one it creates and something messes up the extras). I'll try to find a way to ensure it's initialized, but so far nothing worked.

PR #270, opened 2024-09-09
Switch PHP's default LSP to intelephense instead of phpactor

tonysm:
### Changelog

- Changes the LSP to intelephense instead of the default (phpactor) - intelephense is much less verbose, IMO, it's the one I prefer using myself and also the one I use in VSCode (there's a comparison [here](https://yarnaudov.com/phpactor-vs-intelephense.html)), I made a patch to LazyVim to support it (https://github.com/LazyVim/LazyVim/pull/3691)

---

I started this work at https://github.com/basecamp/omakub/pull/156, but it was mixed with https://github.com/basecamp/omakub/pull/269, so I've decided to send two separate PRs


dhh:
Appreciate the preference, but we're trying to stick to the defaults as much as possible. I would advocate for this change upstream, and Omakub will pick it up if it's changed there.

PR #277, opened 2024-09-19
Headless installation patch

SagaciousFish:
Major changes:
1. Fixed install order on Pure Terminal Installation (`Gum` should now be correctly installed before install/terminal/select-dev-*.sh scripts)
2. Fixed a potential missing requirement for installing 3rd party apps (`apt-add-repository` is now ensured to exist for install/terminal/app-fastfetch.sh)
3. Replacing `bc` dependency with `python3` dependency for version confirming (As `bc` was missing in my test environment, and now `python3` is required by grub2, this should be a safe dependency shift)

Minor fixes:
1. Ensuring `fzf` key-binds script always exists (Otherwise, it might cause an issue after sourcing defaults/bash/init)

Tested environment:
Docker hub image nvidia/cuda:12.6.0-devel-ubuntu24.04 in vscode dev container.

### DIFF
--- a/defaults/bash/shell
+++ b/defaults/bash/shell
@@ -5,6 +5,9 @@ HISTSIZE=32768
 HISTFILESIZE="${HISTSIZE}"
 
 # Autocompletion
+if [ ! -f /usr/share/bash-completion/bash_completion ]; then
+    sudo apt install -y bash-completion
+fi
 source /usr/share/bash-completion/bash_completion
 
 # Set complete path
--- a/install/check-version.sh
+++ b/install/check-version.sh
@@ -9,7 +9,7 @@ fi
 . /etc/os-release
 
 # Check if running on Ubuntu 24.04 or higher
-if [ "$ID" != "ubuntu" ] || [ $(echo "$VERSION_ID >= 24.04" | bc) != 1 ]; then
+if [ "$ID" != "ubuntu" ] || [ $(python3 -c "print(float('$VERSION_ID') >= 24.04)") != 1 ]; then
     echo "$(tput setaf 1)Error: OS requirement not met"
     echo "You are currently running: $ID $VERSION_ID"
     echo "OS required: ubuntu 24.04 or higher"
--- a/install/check-version.sh
+++ b/install/check-version.sh
@@ -9,7 +9,7 @@ fi
 . /etc/os-release
 
 # Check if running on Ubuntu 24.04 or higher
-if [ "$ID" != "ubuntu" ] || [ $(python3 -c "print(float('$VERSION_ID') >= 24.04)") != 1 ]; then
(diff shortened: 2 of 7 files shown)


defaults/bash/shell
```diff
@@ -5,6 +5,9 @@ HISTSIZE=32768
 HISTFILESIZE="${HISTSIZE}"
 
 # Autocompletion
+if [ ! -f /usr/share/bash-completion/bash_completion ]; then
+    sudo apt install -y bash-completion
+fi
```
dhh, line 10: Seems like defensive programming? We're only targeting one distribution. We should know what's installed.

install/check-version.sh
```diff
@@ -9,7 +9,7 @@ fi
 . /etc/os-release
 
 # Check if running on Ubuntu 24.04 or higher
-if [ "$ID" != "ubuntu" ] || [ $(echo "$VERSION_ID >= 24.04" | bc) != 1 ]; then
+if [ "$ID" != "ubuntu" ] || [ $(python3 -c "print(int(float('$VERSION_ID') >= 24.04))") != 1 ]; then
```
dhh, line 12: What exactly does this protect against? It's significantly more complicated.

install/terminal/app-fzf.sh
```diff
@@ -0,0 +1,6 @@
+# Install fzf with key-bindings ensured
+sudo apt install -y fzf
+
+if [[ ! -f /usr/share/doc/fzf/examples/key-bindings.bash ]]; then
+  sudo curl https://raw.githubusercontent.com/junegunn/fzf/master/shell/key-bindings.bash -o /usr/share/doc/fzf/examples/key-bindings.bash
```
dhh, line 5: Seems brittle to depend on some raw gh content dump here.

dhh:
After full review, I don't think I see enough improvements to warrant the changes. Thanks for the suggestions though!

defaults/bash/shell
SagaciousFish: bash-completion is a prerequisite for other potential installs. While it should be installed in vanilla Ubuntu 24.04, it's not always the case for headless images. This is a guarantee for avoiding errors during installing other software.

install/check-version.sh
SagaciousFish: Again, `bc` is not always available, while `Python3` is a prerequisite for a handful of other tools, I assumed python is a safer tool for this task.

install/terminal/app-fzf.sh
SagaciousFish: Agreed, but this is the most official source I could find for `key-bindings.bash`, again for dependencies safe.

SagaciousFish:
Thank you for your review.

PR #278, opened 2024-09-20
feat: additional error handling safety to install.sh script

bradburm:
This update enhances the robustness of the installation script by replacing `set -e` with `set -euo pipefail`. This change ensures:

- The script exits immediately if a command fails (`-e`),
- Any attempt to use an unset variable is treated as an error (`-u`),
- Errors in any part of a pipeline are detected and handled properly (`pipefail`).
- This adjustment provides stronger safeguards and prevents potential issues during system configuration.

Appreciate there are no pipelines in this script, but taking the chance to add it now as there could be in future.

Please review and let me know if any further adjustments are needed. Thank you!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,5 @@
-# Exit immediately if a command exits with a non-zero status
-set -e
+# Exit immediately if a command fails, treat unset variables as an error and ensure all parts of pipelines fail properly
+set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


dhh:
I don't think we actually want any unset variable to cause a halt.  And given that we don't have any pipelines, I also prefer not to program for the future. Appreciate the suggestion, though!

PR #280, opened 2024-09-23
Omakub Mac

tidux:
All the basic tools are available, including iterm2 over Alacritty for better OS integration.  Similarly this also includes tmux for enabling `tmux -CC` / iterm2 integration.

**TODO**: integrate with regular Omakub configurations


dhh:
Thanks for exploring this but Omakub is meant to exclusively be for Ubuntu. No interest in supporting Darwin. But feel free to be inspired and take the ideas to a project that does that.

PR #282, opened 2024-09-24
Adding xclip for allowing nvim to copy to the system clipboard

horvay:
Right now if you try to copy something to the system clipboard in neovim, you get an error 
> 650 lines yanked into "+ <br> clipboard: No provider. Try ":checkhealth" or ":h clipboard".

In my case I was trynig to copy the entire file with gg"+yG (goto top and copy to bottom into system clipboard)

After installing xclip, it worked for me.


dhh:
We already install wl-clipboard for this.

horvay:
Huh, I tried to do `"+y` and nvim literally told me nothing was installed to handle it. Weird because I see the wl-copy utility is installed, but somehow it didn't link up on my system... 😄

Probably just something funky on my install.

PR #297, opened 2024-10-07
fix/sudo-apt-install-sudo

asmrtfm:
remove unnecessary package from prerequisites installation step


Kasui92:
This package is very useful when installing Omakub in containers. In fact, it is often missing here and it is useful to have it.

So, I don't think it's really unnecessary.

asmrtfm:
`sudo` was the package removed because the line `sudo apt install sudo` is well... unnecessary.
If sudo wasn't on the system the that line would not install it, it would throw an error.

PR #298, opened 2024-10-07
adjust sleep-prevention strategy

asmrtfm:
[UPDATED]
Removes redundant idle and lock settings changes from install/desktop.sh and moves the restoring of said settings into install.sh

(Explanation):
install.sh disables sleep by overwriting the session idle delay time and screensaver lock screen settings.
If the desktop environment is Gnome, it calls install/desktop.sh which begins with the exact same lines for disabling sleep.
No other file in the codebase directly calls install/desktop.sh and so there is no reason to repeat these lines - if install/desktop.sh is being ran then sleep will have already be disabled.
However, install/desktop.sh was the only place where those settings were being re-enabled; and since it only gets ran if the desktop environment is Gnome, it made sense just move that part out to install.sh so that both paths (desktop & terminal) could have their settings re-enabled.

I also consolidated the double if conditional and the Gnome check into a single if/else.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -8,6 +8,10 @@ RUNNING_GNOME=$([[ "$XDG_CURRENT_DESKTOP" == *"GNOME"* ]] && echo true || echo f
 source ~/.local/share/omakub/install/check-version.sh
 
 if $RUNNING_GNOME; then
+  # Try to respect any previously configured settings
+  previous_screensaver_lock=$(gsettings get org.gnome.desktop.screensaver lock-enabled) || true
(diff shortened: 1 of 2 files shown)


asmrtfm:
really, the entire contents of install/desktop.sh could be moved into install.sh
but I didn't want to disrupt some as-yet-to-be implemented functionality which may be intended to benefit from their separation.

dhh:
Omakub is only for fresh installations. So no need to try to remember some alternative setting. But we should get rid of the duplication.

asmrtfm:
> Omakub is only for fresh installations. So no need to try to remember some alternative setting. But we should get rid of the duplication.

Done.

PR #299, opened 2024-10-08
Use FZF to select theme

dgmora:
I have a slightly different theme script, but it should work with Omakub too. It looks mostly the same and it lets you type.

Screenshot 2024-10-08 at 18 58 19
Screenshot 2024-10-08 at 18 58 26

I am also using `fd` so that I don't need to add the themes in the script. I'm just showing the folder name though (`rose-pine` instead of `Rose Pine`). Happy to add it here.

```bash
fd --max-depth 1 . "$OMAKUB_PATH/themes" -X basename {} \;
```


dhh:
We use gum everywhere else. I don't think this is something you need to do frequently enough that it's worth diverging from that setup. Appreciate the suggestion though!

PR #307, opened 2024-10-15
Create ~/.local/share/fonts if does not exist

brablc:
On fresh Ubuntu 24.04 this directory does not exist and font installation fails without any error message


dhh:
This is already created in https://github.com/basecamp/omakub/blob/master/install/desktop/fonts.sh#L1 which is run on the first install.

brablc:
Deleting whole .local/share/omakub and starting again solved all my problems on fresh Ubuntu installation.

PR #308, opened 2024-10-17
Support to non-desktop envs to install languages/dbs

gtkatakura:
The **select-dev-language.sh** and **select-dev-storage.sh** scripts require the **gum** package to work. Additionally, we are not displaying the initial choices for non-desktop environments today.


dhh:
Need to point this merge to main rather than stable.

PR #310, opened 2024-10-18
Update app-steam.sh to install 32bit architecture which steam needs

horvay:
On my install, steam did not work without these 32 bit libraries. Is this true for everyone? I can't see how it wouldn't be.

If it's something weird with my install only, sorry for wasting time!

### DIFF
--- a/install/desktop/optional/app-steam.sh
+++ b/install/desktop/optional/app-steam.sh
@@ -1,5 +1,10 @@
 # Play games from https://store.steampowered.com/
 cd /tmp
+
+sudo dpkg --add-architecture i386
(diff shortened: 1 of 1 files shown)


horvay:
To be clear, it needed these libraries at runtime, not at install time.

dhh:
I have not seen a problem getting steam to run on a new install without these. What's the errors you were seeing? What computer?

horvay:
This was a very long time ago. I don't really remember.

It was for my home-built desktop running Ubuntu 24.

I'm going to close this and someone can reopen if an issue arises.

PR #322, opened 2024-10-22
feat/firefox-devedition-installer

asmrtfm:
Adds optional app installer for Firefox Developer Edition.

1) configures mozilla's apt repository
2) if script is ran interactively, it asks the user if they would like to remove the default snap-based firefox
3) finally, it updates and installs firefox-devedition


Babu464-ks (approved):

Jhonatanmizu (approved):

Babu464-ks (approved):

dhh:
Don't think this is a common enough option that it warrants inclusion. Firefox is already part of base Ubuntu.

PR #325, opened 2024-10-27
Text scaling config file and Lenovo profile

jlenriquez:
Changed the config file font size for Alacritty and added new profile that I have tested for  LENOVO laptop using 2.2K display.


dhh:
I think we need to document this instead, as I don't have this hardware to test with, and ensure that it stays in sync. Feel free to suggest something we could add to the manual.

jlenriquez:
Thanks @dhh , it would be great if we could put additional section in the manual re: profiles, we can probably consolidate all the information in that one place. Is the https://manual.omakub.org/ open for contribution as well? I think the more machine profiles we can put in, the more confidence adopting users can also confirm that it would work on their base hardware models.

Issue #226, opened 2024-07-22
Consider adding Zed 


atohovitis:
Zed is now available on Linux. A beautiful new editor built for blazing performance and font rendering quality. 

https://zed.dev/blog/zed-on-linux

rmacklin:
See https://github.com/basecamp/omakub/commit/66182b859df26fca06338d604bcbcd10728c11da

florentdestremau:
There are missing themes for zed to be in sync with theme switching, tho, if you want to contribute 😉

dhh:
Yes, would be great to get Zed themes synced up, if that's possible. But Zed is indeed part of the optional packages now.

PR #327, opened 2024-10-29
Add Zed theme and font synchronization with Omakub

imoneo:
This PR enhances Zed code editor with theme and font switching sync and addresses past issue discussion #226 . 

The synchronization  works both for an already installed Zed configuration and for a fresh installation.

### Notice:
Zed installs extensions at runtime and recognizes them at startup. After the first theme switch, Zed might not change theme without restarting application (if the theme was not  installed before). Only _Rose Pine_ works "out of the box", because this theme is a native Zed extension. This also applies to fonts.


Jhonatanmizu (approved):

dhh:
Thanks for the suggestion, but I don't want to commit to managing configs and themes for any editor except for the two defaults of VSCode and Neovim.

PR #352, opened 2024-12-05
add: improve qol while navigating between workspaces

victor00:
Enable `focus-mode: mouse` for improved usability while navigating between workspaces. It will be "out of the box"  a seamless workflow without clicking to activate a window.


dhh:
Respect this as a choice, but it wouldn't be my default. Thanks for the suggestion!

PR #364, opened 2025-01-20
Add Iosevka Font as an option

alanmaciel:
Add Iosevka font as another programming font option.


dhh:
Appreciate this, but intentionally did not include all Nerd fonts. There are quite a bit more than this as well. The default selection is my personal preferences from what's there.

PR #365, opened 2025-01-21
Resolved the lazyvim checkhealth warnings for luarocks and tree-sitter

hangodek:
I've seen some warnings upon fresh install of lazyvim, installing luarocks and tree-sitter-cli resolve this problems.


dhh:
What are these warnings actually preventing/warning about as a practical matter?

kevinrobayna:
> What are these warnings actually preventing/warning about as a practical matter?

Luarocks is a different way of installing dependencies, not required but they say it's more reliable.

And the treesitter one tbh i have it installed too and i just assume its to have the binary installed and being able to use it so that the plugins behave better (treesitter is only distributed as binary so not sure why it isn't an error tbh)

PR #371, opened 2025-01-26
Add doom-emacs themes

afonsomm:
Add doom-emacs theme synchronization.


dhh:
Appreciate this, but Omakub will only do full configuration for Neovim + VSCode. Don't want to carry configuration code for all other editors.

PR #375, opened 2025-01-30
Fix Neovim installation

shahwan42:
- Update download link
- Update extraction directory name


shahwan42:
I found it's already done here https://github.com/basecamp/omakub/pull/374

Issue #273, opened 2024-09-14
Set default terminal to alacritty


stevenhuyn:
Is useful when using the file explorer and right click contexts of `open terminal here` Also will update `Ctrl + Alt + T` behaviour.

I had to run `sudo update-alternatives --config x-terminal-emulator` to update the default terminal.

Kasui92:
> We already have hotkeys for alacritty, so I'm not sure overwriting the default here is necessary. And I've found that it's actually nice to have a way to use the system terminal in case there are any issues with alacritty.

[#129](https://github.com/basecamp/omakub/pull/129#issuecomment-2254174681)

dhh:
I would actually reconsider #129 if we could make it work with the "Open in Terminal" from the Gnome finder. But the patch there doesn't do that. It only changes `Ctrl + Alt + T`.

iajrz:
After a bit of digging, nautilus seems to have predetermined terminal emulators it's happy to call. I got to [this apparent location](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdesktopappinfo.c#L2701) of the list via [this stack overflow answer](https://askubuntu.com/questions/76712/setting-nautilus-open-terminal-to-launch-terminator-rather-than-gnome-terminal) which is now pointing at the wrong line in the file -- expected, as it's pointing at main, not at a specific commit. 

In any case, the workaround detailed in the linked StackOverflow answer did not work for me, and neither did a few other alternatives I tried.

There might be something that can be done if [this SO answer](https://askubuntu.com/questions/21953/how-do-i-customize-the-context-menu-in-nautilus/889974#889974) and [this SO answer](https://askubuntu.com/questions/827838/can-i-remove-the-open-in-terminal-menu-item-from-nautilus) work.

dhh:
Fixed via #395.

PR #392, opened 2025-02-11
Make Alacritty the default terminal emulator

iajrz:
This makes it work with the "Ctrl + Alt + T" command, as well as any other time a generic terminal emulator is invoked.


iajrz:
This would address #273 in an amiable way.

iajrz:
Noticed it doesn't work with finder's "Open In Terminal" as noted in 273; withdrawing.

PR #395, opened 2025-02-14
Better Alacritty Integration

iajrz:
- Set alacritty as the preferred x terminal emulator, so that e.g. Ctrl+Alt+T opens Alacritty
- Install the nautilus-python extension and a script that results in  adding an "Open Alacritty" option to the Nautilus File Manager context menu

This PR addresses #273 in a thorough manner.

The only change left to be truly opinionated about the terminal would be to find a surgical way to remove the gnome-terminal integration with nautilus which creates its contextual menu option or uninstall it outright.

An additional consideration is to run `nautilus -q` after the extension is installed: the extension does not go live until the file manager is restarted.


dhh:
Don't think we need `nautilus -q` since we're rebooting after the initial setup anyway.

Would be nice to follow this up with a migration that adds it to existing installations. There you might need to reboot!

PR #396, opened 2025-02-17
feat: add ghostty

ofrades:
not sure what people think but here are some changes (two issues pending) to replace alacritty with ghostty #357
credits to @brsc2909 https://github.com/brsc2909/omakub-fedora
- [ ] reload the terminal when changing themes or fonts
- [ ] fix config-file issue
image


dhh:
Not interested in taking on this weight for another terminal. Omakub is built around Alacritty, and I remain happy with that choice.

I'd be OK with Ghostty getting added as an optional install alternative, though. But don't want to carry all the customization integration.

PR #398, opened 2025-02-22
Add umask to bashrc

Roman-Weisman:
Changing to 755, 644 permissions for directories and files respectively.


dhh:
Prefer to stick with the Ubuntu default.

PR #401, opened 2025-02-23
feat: install gum manually with out using apt

billybonks:
The main benefit of using apt is that you can upgrade etc, but because we are manually fetching the package and pushing it into the install feature, you cant have apt do the upgrades later anyway.

I guess one reason to keep it the way that it is, is if it handles variations in CPU architectures as this package is an x86 binary.

The reason for this change is i am making a fork for fedora and the would like to keep as much of the code the same were possible

### DIFF
--- a/install/terminal/required/app-gum.sh
+++ b/install/terminal/required/app-gum.sh
@@ -1,7 +1,8 @@
 # Gum is used for the Omakub commands for tailoring Omakub after the initial install
 cd /tmp
 GUM_VERSION="0.14.3" # Use known good version
-wget -qO gum.deb "https://github.com/charmbracelet/gum/releases/download/v${GUM_VERSION}/gum_${GUM_VERSION}_amd64.deb"
-sudo apt-get install -y ./gum.deb
-rm gum.deb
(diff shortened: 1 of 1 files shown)


billybonks:
alternatively would be to do the install through apt completely

```
sudo mkdir -p /etc/apt/keyrings
curl -fsSL https://repo.charm.sh/apt/gpg.key | sudo gpg --dearmor -o /etc/apt/keyrings/charm.gpg
echo "deb [signed-by=/etc/apt/keyrings/charm.gpg] https://repo.charm.sh/apt/ * *" | sudo tee /etc/apt/sources.list.d/charm.list
sudo apt update && sudo apt install gum
```

dhh:
I don't think there's a win here switching away from using apt on the explicit package downloaded, and I'm not keen to keep expanding the sources list, since it'll break apt update whenever any of them have an issue. Also, gum is mainly just intended for the installer script itself. So appreciate this, but going to skip.

billybonks:
@dhh that reasoning makes sense but in light of your thoughts

>  Also, gum is mainly just intended for the installer script itself.

thoughts on deleting gum when the installation is complete then we do not leave random packages in the computer that are not expected to be used for anything?

PR #402, opened 2025-03-02
fix: curl unzip are used in desktop as well

billybonks:
currently the curl and unzip packages are installed in terminal.sh, but they are used in desktop installers as well, so i combined them with the git as a dependency of the entire script. 
I noticed that we did `sudo apt upgrade -y` in terminal.sh, so moved that as well

image
image

Turns out that unzip is only used in desktop but installed in terminal.sh, if i searched correctly :)


dhh:
Terminal will always be installed, even on desktop. Only reason they're separated is such that a terminal installation doesn't require a desktop part too.

PR #419, opened 2025-03-26
Add full Fedora support for omakub

hangodek:
Tested on both Ubuntu and Fedora distro through VMWare, and both of them work flawlessly


dhh:
Appreciate the work, but I don't want to commit to Fedora support in Omakub, as it would be a maintenance burden going forward. You're free to create a fork of Omakub for Fedora, though!

Josh-Voyles:
@hangodek You're welcome to help me work on my fork. I'm building a Fedora only version to my spec.

PR #421, opened 2025-03-28
Prevent Alacritty from closing on Zellij session detach

### LINKED ISSUE #420: Alacritty closes on Zellij detach
When Alacritty is configured to launch Zellij directly as its shell, the Alacritty window closes once you detach from Zellij. This prevents re-attaching to existing sessions.

Alacritty config file: https://github.com/basecamp/omakub/blob/master/defaults/alacritty.toml

Current config snippet:
```
[shell]
program = "zellij"
```

Problem:
Because Alacritty’s "shell" process exits as soon as Zellij exits, Alacritty closes entirely. This makes it impossible to detach and then re-attach to older Zellij sessions.

Propose config snippet:
```
[shell]
program = "/bin/bash"
args = ["-c", "zellij; exec bash"]
```

Explanation:
- We launch a real shell first (bash here, but any shell works).
- That shell immediately starts Zellij.
- When Zellij exits or is detached, it falls back to Bash instead of closing Alacritty.
- This allows attaching to different Zellij sessions without losing the Alacritty window.

Clay26:
Closes #420 

This tweak still launches Zellij automatically in Alacritty, but it won’t auto-close if you detach. Instead Zellij is launched through bash, so when you detach, you drop back into the shell instead of losing the whole terminal. Now after you restart your computer, you can re-attach to your previous Zellij session.

### DIFF
--- a/defaults/alacritty.toml
+++ b/defaults/alacritty.toml
@@ -2,7 +2,8 @@
 TERM = "xterm-256color"
 
 [shell]
-program = "zellij"
+program = "/bin/bash"
+args = ["-c", "zellij; exec bash"]
 
 [window]
 padding.x = 16


dhh:
I don't think this is necessary? You can do `ctrl o w` to open the session manager in Zellij. From there you can attach to any previous session?

Clay26:
TIL, I didn't realize there was a built in session manager UI. I'll close the PR and the issue. Thanks!

PR #431, opened 2025-04-12
A Port to Fedora 41

renatovico:
This is not a complete fork and need some more love in desktop optional and theme, but the terminal apps already working.

What a learn, the major difference it's the package name and some already is in default for fedora.

Another major difference it's in fzf package, need provide a default keybind.

My idea it's transform to support the both system, but for now the pr it's only reference for another people interesting in fedora too


dhh:
Great to see this! But I don't want to take on the burden of maintaining compatibility for two separate distributions. Feel free to fork this into something you maintain for Fedora, though!

PR #432, opened 2025-04-13
Allow enabling the LazyVIM extras module when installing languages

tonysm:
### Changed

- Another attempt at dynamically enabling the extra modules per-language when installing (see https://github.com/basecamp/omakub/pull/269). This will create the `lazyvim.json` when it doesn't exist and add the language modules (if confirmed).
- It also removes the extra modules when uninstalling the languages

---

This needs https://github.com/LazyVim/LazyVim/pull/5941 to be merged and tagged in LazyVIM. The migration script is currently messing up the modules when it starts for the first time.

### DIFF
--- a/install/terminal/select-dev-language.sh
+++ b/install/terminal/select-dev-language.sh
@@ -6,6 +6,27 @@ else
   languages=$(gum choose "${AVAILABLE_LANGUAGES[@]}" --no-limit --height 10 --header "Select programming languages")
 fi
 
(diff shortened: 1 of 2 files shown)


Kasui92:
UGH! I was hoping for this PR.

tonysm:
@Kasui92 I figured it's just as easy to enable the extras via LazyVim's extras UI... not sure.

PR #447, opened 2025-05-05
added rose pine moon theme

shivamchhuneja:
Added Rose Pine Moon theme.

Background image from Freepik (following the other images.)


bin/omakub-sub/theme.sh
Kasui92, line 1: I think there's a " missing at the end, it doesn't close properly.

shivamchhuneja: Yes, my bad, checked and fixed and tested

.DS_Store
Kasui92, line 1: Remove .DS_Store from the PR, pls. 
I think in this script they are not needed.

shivamchhuneja: Done, thank you for flagging this, added exclusions on my end as well

Kasui92 (approved):

Kasui92:
In #451  I am updating Zellij themes with the new specs.

In case it gets merged, I think this theme could be updated too. There you can find a link to a PR that updates `rose-pine` themes and there is also one for the `moon` variant.

shivamchhuneja:
Just checked your PR as well, will keep an eye on that one and as soon as we move out of the legacy in Zellij, I'll update this one too.

dhh:
Appreciate the effort, but I looked at this when originally setting up the themes, and I don't think it offers enough variance over tokyo night and catpuccin to warrant inclusion by default.

shivamchhuneja:
Noted and a valid point

PR #448, opened 2025-05-05
Improved nvim checks

Kasui92:
I noticed that when I update nvim from Omakub TUI it still regenerates the desktop app from the package.
I think it's okay to remove it.

Also, I only update the nvim theme if it is actually installed.

This PR implements #443  and improves the git history.

### DIFF
--- a/install/terminal/app-neovim.sh
+++ b/install/terminal/app-neovim.sh
@@ -34,5 +34,6 @@ fi
 # Replace desktop launcher with one running inside Alacritty
 if [[ -d ~/.local/share/applications ]]; then
   sudo rm -rf /usr/share/applications/nvim.desktop
+  sudo rm -rf /usr/local/share/applications/nvim.desktop
   source ~/.local/share/omakub/applications/Neovim.sh
 fi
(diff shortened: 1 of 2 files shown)


install/terminal/app-neovim.sh
```diff
@@ -34,5 +34,6 @@ fi
 # Replace desktop launcher with one running inside Alacritty
 if [[ -d ~/.local/share/applications ]]; then
   sudo rm -rf /usr/share/applications/nvim.desktop
+  sudo rm -rf /usr/local/share/applications/nvim.desktop
```
dhh, line 37: I don't have this:

```
ls /usr/local/share/applications/nvim.desktop
"/usr/local/share/applications/nvim.desktop": No such file or directory (os error 2)
```

Kasui92: Normally it is not there, but it may be created in case of a new installation of `neovim`.

Kasui92:
@dhh 

Schermata da 2025-05-19 22-32-24

In the screenshot above, tested on a fresh install of Omakub on Ubuntu 25.04, I ran the `Neovim` update and this happened to me. I then cleaned everything and removed `Neovim` from uninstaller and reinstalled again and the situation returned to me as before. In another installation, however, it did not happen to me.

Here is a screenshot with the generated file.

Schermata da 2025-05-19 22-43-59

I think it could be useful to remove from there too just to be safe.

PR #449, opened 2025-05-05
Add MSCore Fonts 

Kasui92:
It's very convenient to have this font library installed as well.


dhh:
How are they used in LibreOffice? I don't have these installed and defaults seem fine?

Kasui92:
> How are they used in LibreOffice? I don't have these installed and defaults seem fine?

If I'm not mistaken LibreOffice uses Liberation fonts which are supposed to be similar, but not exactly the same visually as MSCore fonts (such as Arial, Verdana, Times New Roman). So yes, LibreOffice works perfectly, but they are not exactly the same fonts.

They are also useful to have just for web browsing. Not that they are very popular fonts nowadays, but I think they can't do any harm.

dhh:
Appreciate the suggestion, but I don't think these are a necessary default, so going to skip for now. Easy for someone to add later.

Issue #450, opened 2025-05-07
Impossible to see selected text on terminal in Kanagawa theme


mdcarreira:
In Kanagawa theme, when text is selected, the theme only changes its color to white. If the original text is already white, the selection becomes invisible.

For example:

![Image](https://github.com/user-attachments/assets/61ce4b79-5d1b-4f70-bd61-d4d48f2a211b)

![Image](https://github.com/user-attachments/assets/e2b8bff3-ace0-4958-9074-81ae0decd65c)

As you can see on the last few lines, unless the text has a color other than white, it's impossible to see the selection.

I recommend setting a light gray background for selected text to make it visible regardless of text color.

Kasui92:
Actually, the selection (and copy) via mouse is not highlighted for Kanagawa... as well as for any theme except for the white text.

I went to the [zellij documentation](https://zellij.dev/documentation/themes.html) to understand how to do it and I discovered, looking at the repository, that they actually changed the way of managing the themes from [this way](https://github.com/zellij-org/zellij/commit/46dae69f567c3badc6e0805250d8f55b61f5f2f0) (now legacy for them) to [this one](https://github.com/zellij-org/zellij/blob/main/zellij-utils/assets/themes/kanagawa.kdl).

Using these styles, the highlighted texts appear.

Maybe we can think about a migration of the zellij themes?

mdcarreira:
Thanks for your finding @Kasui92! I changed my `~./config/zellij/themes/kanagawa.kdl` to the new theme definition you pointed out and it worked like a charm! 

![Image](https://github.com/user-attachments/assets/449b9afd-b27a-46f1-845b-7ef7bd226999)

Kasui92:
You're welcome! 
I'm preparing a PR to fix all the themes

mdcarreira:
Sweet! Feel free to close this issue when you are done 😉

PR #451, opened 2025-05-08
Updated Zellij themes with new specs

Kasui92:
This update is based on [this issue](https://github.com/basecamp/omakub/issues/450) which reports that text highlighting with the mouse no longer appears in the terminal.

Checking the [Zellij documentation](https://zellij.dev/documentation/themes.html) they changed the specs for the themes and those in the previous format are considered [_legacy_](https://zellij.dev/documentation/legacy-themes.html) although they are still supported.

I then proceeded with the alignment and then provided a migration script.

I found all the [themes](https://github.com/zellij-org/zellij/tree/main/zellij-utils/assets/themes) in the Zellij repository, instead for `rose-pine` I relied on [this repository](https://github.com/rose-pine/zellij) and specifically on [this PR](https://github.com/rose-pine/zellij/pull/5), with the update already.

### DIFF
--- a/migrations/1746707329.sh
+++ b/migrations/1746707329.sh
@@ -1,0 +1,17 @@
+# Overwrite old zellij themes with new ones
+if [ -d ~/.config/zellij/themes ]; then
+    files=()
+    for file in ~/.config/zellij/themes/*.kdl; do
+        if [[ -f "$file" ]]; then
+            files+=("$file")
+        fi
+    done
+
+    if [ ${#files[@]} -gt 0 ]; then
+        for file in "${files[@]}"; do
+            if [ -d "$OMAKUB_PATH/themes/$(basename "$file" .kdl)" ]; then
+                cp "$OMAKUB_PATH/themes/$(basename "$file" .kdl)"/zellij.kdl ~/.config/zellij/themes/$(basename "$file" .kdl).kdl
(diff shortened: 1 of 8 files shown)


dhh:
Reverted in https://github.com/basecamp/omakub/pull/456 with the error message this produced.

Kasui92:
> Reverted in #456 with the error message this produced.

The error seems to be due to the use of an old version of Zellij and therefore still expects the legacy format ([ref](https://github.com/zellij-org/zellij/issues/4075)). The new format is used since [version 0.42](https://github.com/zellij-org/zellij/issues/4075) and makes the legacy format incompatible with all functions (such as the text highlighter).

So those who have not updated Zellij can use the legacy, while those who have installed Omakub since March or have updated it should use the new one.

dhh:
Hmm, I saw this with zellij 0.42.2.

imsnif:
Hey - I hope it's alright to chime in. It would be cool to see Omakub use the new theme spec (it solves a lot of issues, including but not limited to the text selection contrast).

I glanced briefly at the PR and don't see any issues with it. Would you maybe like to tell me what you're doing to produce this error so I can take a closer look?

Kasui92:
Fix the errors found in #469

PR #454, opened 2025-05-14
Update Btop/Activity theme

Kasui92:
Just to extend the theme consistency to this panel too.

Most of the themes were retrieved from the same [`Btop` repository](https://github.com/aristocratos/btop/tree/main/themes), while the others from the [`catppuccin`](https://github.com/catppuccin/btop) and [`rose-pine`](https://github.com/rose-pine/btop) repositories.

Here an example with `tokyo-night`

image

### DIFF
--- a/bin/omakub-sub/theme.sh
+++ b/bin/omakub-sub/theme.sh
@@ -7,6 +7,13 @@ if [ -n "$THEME" ] && [ "$THEME" != "<<-back" ]; then
   sed -i "s/theme \".*\"/theme \"$THEME\"/g" ~/.config/zellij/config.kdl
   cp $OMAKUB_PATH/themes/$THEME/neovim.lua ~/.config/nvim/lua/plugins/theme.lua
 
+  if [ -f "$OMAKUB_PATH/themes/$THEME/btop.theme" ]; then
+    cp $OMAKUB_PATH/themes/$THEME/btop.theme ~/.config/btop/themes/$THEME.theme
+    sed -i "s/color_theme = \".*\"/color_theme = \"$THEME\"/g" ~/.config/btop/btop.conf
+  else
+    sed -i "s/color_theme = \".*\"/color_theme = \"Default\"/g" ~/.config/btop/btop.conf
+  fi
+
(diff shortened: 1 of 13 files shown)


Kasui92:
I noticed that `btop` does not always install its **conf file** (https://github.com/Kasui92/omarell/issues/12) and consequently changing the theme via `sed` fails.
I forced its creation during installation and updated the related scripts.

dhh:
This is great! I didn't know that btop could be styled. But it seems like only Tokyo Night looks perfect. It's using the same bg color all the way:

Screenshot from 2025-05-18 18-56-19

Whereas the other themes have different inner background and outer background:

Screenshot from 2025-05-18 18-56-05

All the themes should share the same background with the padding.

Kasui92:
@dhh 
Yes, I noticed.
I saw that the `btop` styles are already fixed, however to further align we can set

```theme_background = False``` 

in [`btop.conf`](https://github.com/basecamp/omakub/blob/master/configs/btop.conf).

So that the `btop` panel can maintain the transparency of `Alacritty`.

dhh:
Let's try it.

Kasui92:
@dhh 
I tested with all the themes and it actually applies transparency. 
With most of the themes, the change has little impact on the display since the background applied by `Btop` is slightly darker than that of `Alacritty`, but with the _Gruvbox_ theme the difference is more evident.

Without trasparency:

Screenshot from 2025-05-19 09-26-12

With transparency:

Screenshot from 2025-05-19 09-26-36

dhh:
Ah, I think that part was already fixed by the changes I did. I was hoping that it would actually make the whole thing transparent, so you could peak at the background image.

Kasui92:
> Ah, I think that part was already fixed by the changes I did. I was hoping that it would actually make the whole thing transparent, so you could peak at the background image.

Mh, it adds transparency, but obviously it is based on the one set on Alacritty and it might not always be evident in fact.

Even trying it on rose-pine with a more marked background some outlines are visible, but not much.


Screenshot from 2025-05-19 10-42-03

PR #459, opened 2025-05-19
COMPATIBLE_MACHINES.md: User-maintained Omakub compatible laptop list

al3rez:
I've successfully installed and used Ubuntu 24.04.2 LTS with Omakub on an Acer Swift SFG14-73 (BIOS V1.14).

It seems to be well compatible and works with minimal extra setup for the main Omakub features. Basic hardware like Wi-Fi and graphics are functional.

I've added this model to a user-maintained list of compatible laptops in `COMPATIBLE_MACHINES.md` that work out of the box with Ubuntu + Omakub.

It would be great to add this model to any official documentation listing compatible hardware for Omakub as well.

Let me know if you need any more details from my setup.


DouglasdeMoura:
Canonical already has a list of Ubuntu certified hardware: https://ubuntu.com/certified/laptops

al3rez:
> Canonical already has a list of Ubuntu certified hardware: https://ubuntu.com/certified/laptops

This isn't going to be the list of certified / pre-installed Ubuntu by manufactors but instead a list of laptops that are new and aren't tested with Linux or latest kernels yet.

Kasui92:
> > Canonical already has a list of Ubuntu certified hardware: https://ubuntu.com/certified/laptops
> 
> This isn't going to be the list of certified / pre-installed Ubuntu by manufactors but instead a list of laptops that are new and aren't tested with Linux or latest kernels yet.

I think such a list might be more useful if associated with Ubuntu rather than Omakub, as it is not really a derived distro but just a configuration script.

al3rez:
It's Opinionated Ubuntu Setup that works only on Ubuntu + Gnome ATM I mentioned this with @dhh and he liked the idea so I created this PR it's really up to him, I like the idea to come to an script setup and find what laptops are compatible with underlying OS instead of finding it myself.

Kasui92:
That's fine! 👍🏻 
Rather than a .md file to insert in this repository (and that would likely be downloaded during installation) we could prepare a page directly in the manual, [here](https://github.com/basecamp/omakub-site), to link in the README. Just to make it easier to find the list and make it more accessible in the future.

dhh:
On second thought, I think the best way to do this is by starting a Discussion and having folks use that to chime in about compatible systems.

PR #460, opened 2025-05-19
Fix web apps installation and uninstallation

### LINKED ISSUE #458: ChatGPT web apps isn't searchable by Ulancher or ubuntu app search
After I install ChatGPT web app it cannot be found in Ulancher / Ubuntu search

Kasui92:
Install and uninstall procedures were not being invoked correctly due to mismatched cases. I normalized following what was done for the other multiple-select chooses.

Fixes #458

### DIFF
--- a/install/desktop/optional/select-web-apps.sh
+++ b/install/desktop/optional/select-web-apps.sh
@@ -1,24 +1,26 @@
 source $OMAKUB_PATH/defaults/bash/functions
 
 AVAILABLE_WEB_APPS=("Chat GPT" "Google Photos" "Google Contacts" "Tailscale")
-apps=$(gum choose "${AVAILABLE_WEB_APPS[@]}" --no-limit --height 7 --header "Select web apps")
+APPS=$(gum choose "${AVAILABLE_WEB_APPS[@]}" --no-limit --height 7 --header "Select web apps" | tr ' ' '-')
 
-if [[ -n "$apps" ]]; then
-  for app in $apps; do
+if [[ -n "$APPS" ]]; then
+  for APP in $APPS; do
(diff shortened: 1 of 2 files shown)


dhh:
I can't replicate that problem on 1.3.0. The case matching we're doing now matches the individual words in the array as is.

Kasui92:
I also encountered this problem on a fresh installation of v1.3.0 on Ubuntu 25.04.

If you try to uninstall from Omakub panel does it remove them?

dhh:
It does. Odd. Let me try in my VM to confirm.

dhh:
I can't replicate it on my VM either. But I did find that we shouldn't be using --no-sandbox as an option for these web app stubs! So will have to cut a point release for that.

But maybe it's a 25.04 issue? I tested on 24.04. Will try 25.04.

Kasui92:
I'm also trying with Ubuntu 24.04. On Ubuntu 24.10 I had the same problem.

Kasui92:
Ok, I may have figured out the original problem (... _and I'm a little ashamed because I followed it too!_)

I think the reason this anomaly was generated is because it didn't toggle the choices, but just pressed Enter on the Chat GPT option without actually choosing it. By doing this, it doesn't actually take any choices and the array in apps is empty.

If you make the selections instead, the case is matched.

Screenshot From 2025-05-20 12-52-57

So, yes... Everything works, we just didn't use the tool well (me first, sorry!).

However, instead of composing the array with all the words (so both "Chat" and "GPT"), isn't it better to work with complete labels (so "chat-gpt")? 

Also, since there is no option to go back, maybe we can put that and also a warning message where no option is chosen?

PR #461, opened 2025-05-19
Fix PHP installation in no-LTS version

Kasui92:
I did a little research around and indeed the `ppa:ondrej/php` repository **does not support** non-LTS versions of Ubuntu. 

However it's  possible to [force](https://askubuntu.com/questions/1530400/install-older-php-versions-on-24-10) this to use the repository associated with noble in order to make PHP available for other versions as well.

Schermata da 2025-05-19 23-17-38

Fixes this #342

### DIFF
--- a/install/terminal/select-dev-language.sh
+++ b/install/terminal/select-dev-language.sh
@@ -21,6 +21,7 @@ if [[ -n "$languages" ]]; then
       ;;
     PHP)
       sudo add-apt-repository -y ppa:ondrej/php
+      sudo sed -i 's/"$(lsb_release -sc)"/noble/g' /etc/apt/sources.list.d/ondrej-ubuntu-php-"$(lsb_release -sc)".sources
       sudo apt -y install php8.4 php8.4-{curl,apcu,intl,mbstring,opcache,pgsql,mysql,sqlite3,redis,xml,zip}
       php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
       php composer-setup.php --quiet && sudo mv composer.phar /usr/local/bin/composer


Kasui92:
Unfortunately it seems that some libraries for `php8.4-zip` and `php8.4-intl` are not available for Ubuntu25.04. I'll investigate to find a solution.

florentdestremau:
Another solution might be to install php8.4 from the original sources rather than the custom ppa, but this forces you into a php version that may be outdated (like in 24.04 it's php8.3).

Kasui92:
I don't know what changed, but this morning I didn't encounter the extension installation error anymore. I also tested it on two fresh installs of Ubuntu 25.04 and everything worked fine. 🙄

Kasui92:
I don't know what changed, but this morning I didn't encounter the extension installation error anymore. I also tested it on two fresh installs of Ubuntu 25.04 and everything worked fine. 🙄

Screenshot From 2025-05-20 10-25-30

Kasui92:
I should have discovered the reason for the errors I encountered. 
Most of it is related to the removal phase of PHP and the PPA used to obtain the necessary libraries. This is not removed correctly using the usual command, but the [WIKI](https://github.com/oerdnj/deb.sury.org/wiki/How-to-downgrade-to-stock-packages#how-to-downgrade-packages-to-the-pristine-state) itself [recommends using ppa-purge which correctly cleans references and dependencies.](https://github.com/oerdnj/deb.sury.org/wiki/How-to-downgrade-to-stock-packages#ubuntu)

> Recently, there was an [issue](https://github.com/oerdnj/deb.sury.org/issues/1382) about restoring the extra packages to the original state because the repository comes with extra libraries that stop getting the updates when the [deb.sury.org](https://deb.sury.org/) repository is removed from /etc/apt/sources.list.d.

The installation therefore occurs correctly even without further checks, just replace the distro reference with that of Ubuntu 24.04, and in the removal it uses ppa-purge.

dhh:
Do we still need this? Maybe the best would indeed just be to go with the default for the distro?

dhh:
Agree. If you want to do a PR for just going with stock, let's do that.

PR #463, opened 2025-05-21
Record the migrations ran

joel:
The main issue is that when a migration fails, we cannot determine the final state of the machine's configuration. 

We are unable to identify which specific migration caused the failure. 

The migrations are not idempotent, meaning they cannot be replayed safely. 

Any failure requires manual investigation. 

This lack of clarity undermines trust in the process.

This pull request aims to initiate the traceability of executed migrations, but I believe migrations require further consideration.

### DIFF
--- a/bin/omakub-sub/migrate.sh
+++ b/bin/omakub-sub/migrate.sh
@@ -64,7 +64,7 @@ for file in $OMAKUB_PATH/migrations/*.sh; do
 
   # Skip this migration if we shouldn't run it
   if ! $should_run; then
-    echo "Migration $migrate_at skipped: $migration_reason"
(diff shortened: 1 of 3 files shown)


dhh:
Again, I appreciate all the work done here. It's quite a large body! But it's also too much for Omakub directly.

What I'd like to do with migrations is to ensure that they're always idempotent, and then we can add a way to run all migrations from scratch, if there was ever a problem.

joel:
@dhh  Making migration more robust will improve the experience. I installed Omakub three times, and for unknown reasons, I encountered issues twice with migration that forced me to investigate where it failed.

PR #464, opened 2025-05-21
Add Logging Helpers for Improved Traceability

joel:
This pull request introduces a shared logging utility (helpers.sh) for Omakub installation.

Simple functions (setup_log, log_message) that can be reused across all scripts.

Rationale

Making troubleshooting much easier.


nareshrajkumar866 (approved):

joel:
Hi, @nareshrajkumar866, thanks for approving this PR. I removed the app-chrome-beta example. I'm [running an updater in parallel](https://github.com/joel/omakub-supercharged) to keep up with the main project and have the setup I need. If you have anything that interests you, just let me know, and I will open a PR.

nareshrajkumar866 (approved):

bin/omakub
Kasui92, line 1: Once you have added the `shared` folder and the helpers file, you don't need to do this check. 
It makes sense to assume it's there.

dhh:
Appreciate the work here. I know the temptation to extract things out and make it more of a framework! I've attempted it a few times, but then I always come back to wanting to keep this as simple as possible with little if any internal dependencies. Each bash script should be as self-contained as feasible in almost all cases. Even if that means a tad of duplication.

PR #469, opened 2025-05-27
Reapply "Updated Zellij themes with new specs"

Kasui92:
This reverts commit 4e82c7b01287be32cfb4ccc63a51e55397479a77.


Kasui92:
I've been using these settings for several months now and have also tested them on other devices. At this point, I think it was a migration issue.

For now, I'll re-enable it, hoping for better luck. I don't know if Zellij will still be available on Omakub, but with these styles, it's possible to regain control over underlining.

Kasui92:
@dhh 
Updated the main theme schemas, added new ones, and improved the migration script.

PR #472, opened 2025-05-30
Additional Browsers

joel:
Bonus, shows install status

Image


joel:
Need https://github.com/basecamp/omakub/pull/464

PR #478, opened 2025-06-06
Make Alacritty start in Fullscreen mode

brentdodell:
Given that Omakub is workspace focused, and the demo video talks about a "one app per workspace" flow (6:49-7:24), it would seem to make sense to default Alacritty to fullscreen.


dhh:
Appreciate the suggestion, but I think this is too much. I never run full screen apps on my 6K XDR, even if I often do it on the laptop.

PR #479, opened 2025-06-07
Install Pinta via flatpak instead of snap

jjatinggoyal:


dhh:
What's the benefit you've found from using the flatpak version?

Jhonatanmizu (approved):

PR #481, opened 2025-06-09
Add script for uninstalling vscode

jjatinggoyal:
Script for uninstalling vscode was missing. `set-vscode-theme.sh` already checks for presence of vscode before setting themes.

### DIFF
--- a/uninstall/app-vscode.sh
+++ b/uninstall/app-vscode.sh
@@ -1,0 +1,2 @@
+sudo apt purge -y remove code
+rm -rf ~/.config/Code/User


Kasui92:
I did [something similar](https://github.com/Kasui92/omarell/blob/dev/uninstall%2Fapp-visual-studio-code.sh) on my Omakub fork and found it useful to purge instead of just remove and clean up user configurations as well, to make future installations easier.

dhh:
Yeah, we should move everything over to purge 👍

PR #507, opened 2025-07-19
boot.sh: Warning on installations currently not using GNOME

l1ghty:


Kasui92 (approved):
In theory, that's correct. 👍🏻 

The only thing I can think of is that Omakub can currently be installed on Ubuntu Server, which doesn't actually have a DE/GNOME and could therefore generate unnecessary noise. ONLY IN THAT CASE, of course. 

However, it's useful if you install it on older versions of Ubuntu (which wouldn't pass the checks earlier) or on derived distros.

PR #510, opened 2025-07-21
Update app-vscode.sh

neidiom:
I was getting this error:

wget -qO- https://omakub.org/install | bash
________                  __        ___.
\_____  \   _____ _____  |  | ____ _\_ |__
 /   |   \ /     \__   \ |  |/ /  |  \ __ \
/    |    \  Y Y  \/ __ \|    <|  |  / \_\ \
\_______  /__|_|  (____  /__|_ \____/|___  /
        \/      \/     \/     \/         \/

=> Omakub is for fresh Ubuntu 24.04+ installations only!

Begin installation (or abort with ctrl+c)...
E: Conflicting values set for option Signed-By regarding source https://packages.microsoft.com/repos/code/ stable: /etc/apt/keyrings/packages.microsoft.gpg != /usr/share/keyrings/microsoft.gpg E: The list of sources could not be read.

The Problem I Faced

  When I first tried to run the installation, I immediately ran into a persistent error from your system's package manager, apt.

  The core of the problem was a conflict in your software sources. The repository for Microsoft VS Code was defined in more than one place, and these definitions were secured with two different, conflicting GPG keys. This created a
  deadlock; apt would refuse to do anything until the conflict was resolved, which in turn caused the installation script to fail right at the start.

  My initial attempts to fix this weren't working because I was placing the cleanup logic in scripts that were being called too late in the process. The error was happening before my fixes could even run.

  The Change I Made

  After realizing the timing was the issue, I corrected my approach.

  I pinpointed the exact script that was triggering the error: install/desktop/app-vscode.sh. I then modified this file to perform a comprehensive cleanup at the very beginning, before it could do anything else.

  The logic I added to the start of that script now does the following:
   1. It actively searches for and removes any apt source file on your system that mentions the conflicting Microsoft repository.
   2. It explicitly removes both of the GPG keys that were causing the conflict.
   3. Finally, it runs sudo apt-get update to confirm that the fix was successful.

  By placing this robust cleanup process in the correct script, I ensured that the conflict is resolved at the precise moment necessary, allowing the installation to proceed on a clean and stable foundation.


neidiom:
I have taken a more gentle approach.

Kasui92:
@neidiom 
IMO, the script generates too much noise (due to comments, both printed to the terminal and in the script) and performs unnecessary checks/steps.

It's sufficient an initial check that only verifies whether the GPG keys and Microsoft repo have already been added. If so, it continues with the VSCode installation from APT; otherwise, it adds the mirrors first and then continues.

This allows for much less clutter and keeps the code clean.

neidiom:
@Kasui92 please check now. I wrapped the repo setup part into a function.

PR #519, opened 2025-07-29
Enable fingerprint auth PAM in ubuntu

samuelpecher:
In Ubuntu use pam-auth-update to enable the fprintd profile in common-auth PAM to allow fingerprint auth for sudo and applications which hook into PAM-based auth (1password!).

There's a condition to check for existing fingerprints if the user has already enrolled any or if this is an update through the migration framework.

Essential for sanity if coming from alternative systems with pervasive touch identification.

### DIFF
--- a/migrations/1753807915.sh
+++ b/migrations/1753807915.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# Run fingerprint setup
+source ~/.local/share/omakub/install/fingerprint-setup.sh


santiagonbernardes:
I'm having issues using fingerprint on Ubuntu 24.04.2. I can enable and use it but it doesn't unlock the keyring, so I have to do it using my password. The error message is:

`the login keyring did not get unlocked when you logged into your computer fingerprint` 

 Just tought it might useful to let you know.

samuelpecher:
@santiagonbernardes the password is required to decrypt the keyring, so eventually you will need to manually enter the password once per session at least.

PR #546, opened 2025-09-05
fix: OMAKUB_PATH point to wrong path when installed for root

IceCodeNew:


IceCodeNew:
Even though installation for root user is an anti-feature of this project, this change should not hurt anyone.

PR #560, opened 2025-10-29
Add Starship

Kasui92:
This PR adds Starship to Omakub, using Omarchy’s configuration and style as a reference for consistency.

Screenshot From 2025-11-10 00-07-16

### DIFF
--- a/install/terminal/app-starship.sh
+++ b/install/terminal/app-starship.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
-curl -sS https://starship.rs/install.sh | sh -s -- -y
+cd /tmp
+curl -sLo starship.tar.gz "https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-musl.tar.gz"
+tar -xf starship.tar.gz starship
+install starship ~/.local/bin
+rm starship.tar.gz starship
+cd -
 
 cp ~/.local/share/omakub/configs/starship.toml ~/.config/starship.toml
--- a/install/terminal/app-starship.sh
+++ b/install/terminal/app-starship.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+curl -sS https://starship.rs/install.sh | sh -s -- -y
+
+cp ~/.local/share/omakub/configs/starship.toml ~/.config/starship.toml
--- a/bin/omakub-sub/install.sh
+++ b/bin/omakub-sub/install.sh
@@ -18,6 +18,7 @@ CHOICES=(
   "Ollama            Run LLMs, like Meta's Llama3, locally"
   "Retroarch         Play retro games"
   "Spotify           Stream music from the world's most popular service"
+  "Starship          Command line prompt for astronauts"
   "Steam             Play games from Valve's store"
   "Tailscale         Mesh VPN based on WireGuard and with Magic DNS"
   "VirtualBox        Virtual machines to run Windows/Linux"
@@ -27,7 +28,7 @@ CHOICES=(
   "<< Back           "
 )
 
-CHOICE=$(gum choose "${CHOICES[@]}" --height 26 --header "Install application")
+CHOICE=$(gum choose "${CHOICES[@]}" --height 27 --header "Install application")
 
 if [[ "$CHOICE" == "<< Back"* ]] || [[ -z "$CHOICE" ]]; then
   # Don't install anything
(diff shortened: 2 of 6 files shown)


install/terminal/app-starship.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+curl -sS https://starship.rs/install.sh | sh -s -- -y
```
anmolnar: Ubuntu 25.10 already has `starship` in deb [package](https://launchpad.net/ubuntu/+source/starship).

Kasui92: You're right, but that wasn’t the case in earlier releases, and the script still needs to work on 24.04. On top of that, the deb packages are frozen at older versions. 
In this situation, keeping the tool on the latest available release actually makes more sense.

anmolnar: 1.22.1 vs the latest 1.24.1 doesn't seem to the end of the world and I don't see it as frozen. A packaged version always has the benefit of getting updated automatically while if you install from the source, you'll have to keep track of your installed applications yourself. Omakub supports DEB packages, Snap and Flatpak. They might have a suitable package which is well-maintained.

ImmortalSom: > 1.22.1 vs the latest 1.24.1 doesn't seem to the end of the world and I don't see it as frozen. A packaged version always has the benefit of getting updated automatically while if you install from the source, you'll have to keep track of your installed applications yourself. Omakub supports DEB packages, Snap and Flatpak. They might have a suitable package which is well-maintained.

I looked into it, and you're more right. This method of installation, where it “poops” straight into  /usr/local/bin, is a disgusting practice. In fact, it's worth making a script that downloads and places the executable file in ~/.local/bin.

```
cd /tmp
curl -sLo starship.tar.gz "https://github.com/starship/starship/releases/latest/download/starship-x86_64-unknown-linux-musl.tar.gz"
tar -xf starship.tar.gz starship
install -m 755 starship ~/.local/bin
rm starship.tar.gz starship
cd -
```

Kasui92: Seems fair to me. Fixed! 💪🏻

ImmortalSom:
`/defaults/bash/prompt`

It would be worth deleting, as it loses its meaning.
See https://github.com/basecamp/omarchy/pull/3295

> Fix PS1 setting no longer necessary given default starship installation by @dhh

Kasui92:
> `/defaults/bash/prompt`
>
> It would be worth deleting, as it loses its meaning.

I was actually thinking about that, but the way `/defaults/bash` are handled here is different from Omarchy. The risk is that anyone not using Starship could suddenly end up without a prompt. We’d probably need some kind of safeguard to avoid making the change too disruptive.

With #562  we would manage them in a similar way and then maybe do the migration more "gentle".

ImmortalSom:
> The risk is that anyone not using Starship could suddenly end up without a prompt. 

You're mistaken. You're misinterpreting what's happening here. We substitute the system prompt. It remains, it's not going anywhere. If something goes wrong, the system prompt will appear.
This is a continuation of my argument about PS1. I insisted there that if something goes wrong, let the system prompt return.

Kasui92:
Anyone who doesn’t want to use Starship and prefers to stick with the old PS1 setup, with NerdFont and icons, would lose it immediately. That’s exactly what happened in Omarchy at first, because the PS1 wasn’t removed right away after Starship landed. I think the same approach makes sense here: phase it out slowly and give people who don’t want Starship a way to keep their existing prompt.

ImmortalSom:
> Anyone who doesn’t want to use Starship and prefers to stick with the old PS1 setup, with NerdFont and icons, would lose it immediately. That’s exactly what happened in Omarchy at first, because the PS1 wasn’t removed right away after Starship landed. I think the same approach makes sense here: phase it out slowly and give people who don’t want Starship a way to keep their existing prompt.

This is not the way of Omak. 😄

Kasui92:
Yes, of course, it’s definitely less Omakase. But between removing the PS1 entirely and keeping it no matter what (which I don’t really see the point of either), I think there’s room for a middle ground if you want to switch everything in one go.

Maybe we could consider a migration step that adds the current PS1 back into `.bashrc` for anyone who doesn’t want to install Starship. 

Otherwise, fine by me... we can just clean up the PS1 and move on. 👍🏻

PR #562, opened 2025-11-08
Enhance Bash default configuration robustness 

Kasui92:
Following the approach used in Omarchy, reorganized all Bash scripts into `/defaults/` directory, made `.bashrc` easier to modify and introduced preventive checks for aliases relying on packages or libraries that might be uninstalled.


ImmortalSom:
I want to give you some practical advice. In my work, I connect aliases that are clearly linked to applications in this way.

```
# Load custom aliases
if [ -d "$HOME/.config/aliases" ]; then
    for f in "$HOME/.config/aliases/"*.sh; do
        [ -f "$f" ] && source "$f"
    done
fi
```

app-lazygit.sh add

```
    mkdir -p ~/.config/aliases
    printf "%s\n" \
        'alias lzg="lazygit"' |
        tee ~/.config/aliases/lazygit.sh >/dev/null
```

As far as I'm concerned, they should be tied specifically to installing and uninstalling applications.

PR #574, opened 2025-11-16
fix: Ulauncher, add 10s delay

ImmortalSom:
On weak PCs and virtual machines, a delay of 2 seconds is not enough.


Kasui92 (approved):

ImmortalSom:
I accidentally closed this pull request. It's because it's generally structured incorrectly, but I didn't provide an alternative option. However, since it still solves the current problem in its current state, I realized that I was wrong and that it's worth leaving this pull request until a solution to this problem that I consider correct is submitted.

PR #575, opened 2025-11-17
feat: add full Debian 13 support

ImmortalSom:
Screenshot at Nov 17 06-20-14

Everything is working!

Before installing, you must first install sudo from root and add the user to the sudo group.

```
su

apt install -y sudo
sudo adduser <username> sudo
su <username>
```


Kasui92 (commented):
As I mentioned, I don’t think this change really fits an Ubuntu-specific script. It makes a lot more sense as part of a separate project, aimed at Debian and its users.

That said, I get your point about PPAs not always being ideal or recommended. It could still be worth applying your changes where they make sense and suggesting alternatives that don’t rely on PPAs.

ImmortalSom:
To be honest, I see the point in distributing these dotfiles to all possible Linux distributions. The main limitation here is that everything revolves around GNOME. It's not that difficult to set up Omakub for installation in almost all common distributions, even those that aren't based on deb packages.

As for me, the main point of Omakub is the opportunity to start here and now without radically changing anything. For people who are not yet used to working with the console, GNOME remains with all its goodness.

Kasui92:
The main issue is the larger maintenance surface you get when adding Debian. Its lifecycle and packages differ quite a bit from Ubuntu’s. And that’s without even considering Gnome, which Ubuntu customizes more heavily compared to Debian’s vanilla setup. This means maintainers and contributors would have to test every change across multiple environments. I’m already not thrilled about supporting non-LTS Ubuntu releases, and prefer sticking to LTS only. 

So I honestly think it’s better not to merge everything into one project, but to keep separate spin-offs or parallel projects that aim for the same result.

ImmortalSom:
I'm busy with a slightly larger change, but I think I have something to send right now to fix the situation.

ImmortalSom:
Although it works, improvements are needed in this area, which I am successfully implementing in my more extensive work.

robeastham:
@ImmortalSom great to hear that this is working on Debian 13. I'm just about to spin up a test. Did the fixes you made get merged anywhere? Either way do your commits above assume a basic server install of Debian or does one have to install the gnome desktop package before installing Omakub?

PR #577, opened 2025-11-22
Switching to the official repository

ImmortalSom:
This repository is listed as official on the project page:
https://github.com/charmbracelet/gum


ImmortalSom (commented):
Since version 24.04 *.list files are recognized as deprecated, and the approach itself, even with the add-apt-repository command, is also deprecated and unsafe.

Kasui92:
Seems all good to me!

Kasui92 (approved):

PR #578, opened 2025-11-22
refactor(apt): align repository configuration with Ubuntu .sources standard

ImmortalSom:
Migrate PPA and third-party repos to structured .sources format
with explicit Signed-By, Architectures, and proper key handling
in /etc/apt/keyrings/.

P.S. add-apt-repository deprecated. In short, add-apt-repository didn't add keys as separate files, but imported them into the shared keyring. A complete switch to *.sources, with explicit key file specifications, is safer and preferable.

### DIFF
--- a/install/terminal/app-fastfetch.sh
+++ b/install/terminal/app-fastfetch.sh
@@ -5,9 +5,7 @@
 # Load OS release information
 . /etc/os-release
 if [ ! -f /etc/apt/sources.list.d/zhangsongcui3371-fastfetch.sources ]; then
-    sudo mkdir -p /etc/apt/keyrings
-    sudo rm -f /etc/apt/keyrings/zhangsongcui3371-fastfetch.asc
-
+    [ -f /etc/apt/keyrings/zhangsongcui3371-fastfetch.asc ] && sudo rm /etc/apt/keyrings/zhangsongcui3371-fastfetch.asc
     curl -fsSL \
         "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xeb65ee19d802f3eb1a13cfe47e2e5cb4d4865f21" \
         | sudo tee /etc/apt/keyrings/zhangsongcui3371-fastfetch.asc > /dev/null
--- a/install/terminal/app-fastfetch.sh
+++ b/install/terminal/app-fastfetch.sh
@@ -1,7 +1,27 @@
 #!/bin/bash
 
 # Display system information in the terminal
-sudo add-apt-repository -y ppa:zhangsongcui3371/fastfetch
+#
+# Load OS release information
+. /etc/os-release
+if [ ! -f /etc/apt/sources.list.d/zhangsongcui3371-fastfetch.sources ]; then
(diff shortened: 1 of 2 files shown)


install/terminal/app-fastfetch.sh
```diff
@@ -1,7 +1,27 @@
 #!/bin/bash
 
 # Display system information in the terminal
-sudo add-apt-repository -y ppa:zhangsongcui3371/fastfetch
+#
+# Load OS release information
+. /etc/os-release
```
Kasui92: Is it necessary to enter the version codename? Can't it be left blank?

ImmortalSom: There are two things to note. I added this line just so I could run this file by itself for testing. Actually, install.sh already has this call.

Second, I haven't tested it. There's a mechanism where if you specify * it will take the latest version. But something in my subconscious tells me not to do that (just in case, Fastfetch doesn't have any other versions, that's why it has *).

Again, I'm having trouble with this because I'm using "oracular" there in Debian. I wouldn't touch it.

P.S. I'm new to git, and I haven't developed good habits yet, so I'm using it like Ctrl+S, so forgive me.

ImmortalSom: There are two things to note. I added this line just so I could run this file by itself for testing. Actually, install.sh already has this call.

Second, I haven't tested it. There's a mechanism where if you specify * it will take the latest version. But something in my subconscious tells me not to do that (just in case, fastfetch doesn't have any other versions, that's why it has *).

Again, I'm specifically putting "oracular" there for Debian. I wouldn't touch it.

P.S. I'm new to git, and I haven't developed good habits yet, so I'm using it like Ctrl+S, so forgive me.

install/terminal/app-fastfetch.sh
```diff
@@ -1,31 +1,27 @@
...
-
-    printf '%s\n' \
-        "Types: deb deb-src" \
-        "URIs: https://ppa.launchpadcontent.net/zhangsongcui3371/fastfetch/ubuntu" \
-        "Suites: ${VERSION_CODENAME}" \
-        "Components: main" \
-        "Architectures: amd64" \
-        "Signed-By: /etc/apt/keyrings/zhangsongcui3371-fastfetch.asc" \
-        | sudo tee /etc/apt/sources.list.d/zhangsongcui3371-fastfetch.sources > /dev/null
+	sudo apt install -y gnupg
```
Kasui92: This could be put in `/install/terminal.sh`, it might be useful after all.

ImmortalSom:
Essentially, this commit makes no sense for Ubuntu. In fact, I used it myself to add a repository to Debian.

Kasui92 (approved):

Kasui92:
> Essentially, this commit makes no sense for Ubuntu. In fact, I used it myself to add a repository to Debian.

I don't understand which commit you're referring to 🤔

ImmortalSom:
> I don't understand which commit you're referring to 🤔

I got my words mixed up. I meant that this pull request for Ubuntu doesn't make sense.

PR #579, opened 2025-11-24
feat: enhance PS1

ImmortalSom:
- with dynamic Nerd Font icon for GUI terminals, fallback to standard prompt in TTY

### DIFF
--- a/defaults/bash/prompt
+++ b/defaults/bash/prompt
@@ -3,5 +3,10 @@ force_color_prompt=yes
 color_prompt=yes
 
 # Simple prompt with path in the window/pane title and caret for typing line
-PS1=$'\uf0a9 '
-PS1="\[\e]0;\w\a\]$PS1"
+if [ -n "$DISPLAY" ] || [ -n "$WAYLAND_DISPLAY" ]; then
+	# Only change prompt if in graphical mode AND Nerd Font is available
+	if command -v fc-list >/dev/null 2>&1 && fc-list | grep -qi "Nerd Font"; then
+		PS1=$'\uf0a9 '
+		PS1="\[\e]0;\w\a\]$PS1"
+	fi
+fi


Kasui92 (commented):
In my opinion, this moves quite far from the project’s “Omakase” vision. Conceptually it’s fine and it does make things more flexible, but since it would still require users to customize it to handle “unexpected” cases or override the configuration directly, it’s better to keep a clean default setup and focus on making it easy to extend or override through someone’s own bash profiles.

ImmortalSom:
The problem is that something can go wrong during installation, or you can connect to your computer via ssh and see a strange square when you log in as a user. Again, Omakub allows you to install only terminal programs, and in these cases the symbol will be broken everywhere.

This is not an intrusion into Omakub field, it is a patch for exceptional cases.

Kasui92:
> Again, Omakub allows you to install only terminal programs, and in these cases the symbol will be broken everywhere.

Exactly! I think the best solution is to drop the option to install only the terminal setup. I get why it exists, but focusing solely on the full desktop experience would make things much simpler for everyone, users and maintainers.
What do you think about it, @dhh?

dhh:
Yes, happy just to focus on the desktop setup. I never ended up using the terminal-only setup.

ImmortalSom:
I'm probably misunderstanding something, or because I'm using a translator, I'm not getting my point across very well.

This doesn't make any changes to the desktop. It just allows you to show the standard OS prompt for input where it's not possible to show this symbol.

It's just that this symbol is only available in the Nord Font family, and if it's not there or you're not in GUI mode, the input prompt will either be empty or a rectangle, which can be confusing.

Such situations can arise if something went wrong or someone deleted the fonts, or there was a network failure and the fonts were not installed, or you logged into the computer via ssh to do something, or the X/Wayland session crashed, or 100 other reasons. The first time this happened to me, it was exactly one of these reasons, and I stared at a blank screen for several minutes, not understanding where my prompt was.

Kasui92:
I think I understand the reasoning behind the change, and I'm not saying it's wrong. There are several valid scenarios where it could make sense.

That said, I don't really see the need for it. It's meant to be a default provided by the installation, and anyone who wants the previous setup can simply override it. By default, Omakub ships with that font and is expected to render characters and icons correctly. If someone chooses not to use it, fixing the PS1 becomes their own responsibility. I wouldn’t worry about handling fallbacks or trying to cover every error case.

But ... I won’t argue against it either.
As I said, it’s not a meaningless idea. I simply wouldn't bother exploring it.

ImmortalSom:
I was curious about Omarchy. Were there really no complaints there either? It turns out Omarchy installed Starship on August 17th and has no issues in console mode.

Kasui92:
But Omarchy works differently: it uses `uswm` which is like encapsulating the apps in a separate env, while in TTY they have the basic PS1.

PR #591, opened 2025-12-31
fix: use runtime fallback for zoxide cd function

ki11e6:
## Summary
- Removes `alias cd='z'` from `defaults/bash/aliases`
- Adds a `cd()` function in `defaults/bash/init` with **runtime fallback** to `builtin cd`

## Problem
The `cd` command fails with `bash: command not found: __zoxide_z` in non-interactive shell environments (e.g., Claude Code CLI, scripts, CI/CD pipelines).

**Root cause**: Zoxide's `__zoxide_z` function is not available in non-interactive shells because `eval "$(zoxide init bash)"` doesn't properly define functions in these contexts. See: https://github.com/anthropics/claude-code/issues/2407

## Previous Fix Attempt (didn't fully work)
```bash
# Source-time check - only creates alias if function exists at source time
if declare -f __zoxide_z &> /dev/null; then
  alias cd='z'
fi
```

**Why it failed**: This checks only once when the file is sourced. In some environments, the shell state can become inconsistent (functions defined but later unavailable), or tools like `mise` can override the `cd` function after our alias is set.

## New Fix (runtime fallback)
```bash
cd() {
  if declare -f __zoxide_z &>/dev/null; then
    __zoxide_z "$@"
  else
    builtin cd "$@"
  fi
}
```

**Why this works**:
1. Checks at **execution time** whether `__zoxide_z` is available
2. Uses zoxide when available (interactive shells)
3. Falls back gracefully to `builtin cd` when not (non-interactive shells)
4. Never produces errors in any environment

## Test Results
| Environment | cd behavior | Result |
|-------------|-------------|--------|
| Interactive shell | Uses `__zoxide_z` | ✅ zoxide works |
| Non-interactive shell | Falls back to `builtin cd` | ✅ no errors |
| Claude Code CLI | Falls back to `builtin cd` | ✅ no errors |

## Test plan
- [x] Verified `cd` uses zoxide in interactive shells
- [x] Verified `cd` falls back to builtin in non-interactive shells
- [x] Verified no `__zoxide_z` errors in any environment
- [x] Verified `..`, `...` aliases still work in interactive shells

### DIFF
--- a/defaults/bash/aliases
+++ b/defaults/bash/aliases
@@ -5,7 +5,6 @@ alias lt='eza --tree --level=2 --long --icons --git'
 alias lta='lt -a'
 alias ff="fzf --preview 'batcat --style=numbers --color=always {}'"
 alias fd='fdfind'
-alias cd='z'
 
 # Directories
 alias ..='cd ..'
--- a/defaults/bash/init
+++ b/defaults/bash/init
@@ -4,11 +4,16 @@ fi
 
 if command -v zoxide &> /dev/null; then
   eval "$(zoxide init bash)"
-  # Only alias cd to z if __zoxide_z function was properly initialized
-  # This prevents broken alias in non-interactive shells (e.g., Claude Code)
-  if declare -f __zoxide_z &> /dev/null; then
-    alias cd='z'
-  fi
+  # Override cd with a function that has runtime fallback
+  # This handles non-interactive shells (e.g., Claude Code) where zoxide
+  # functions may not be available, falling back to builtin cd gracefully
+  cd() {
+    if declare -f __zoxide_z &>/dev/null; then
+      __zoxide_z "$@"
+    else
+      builtin cd "$@"
+    fi
+  }
 fi
 
 if command -v fzf &> /dev/null; then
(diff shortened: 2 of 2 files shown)


Kasui92 (commented):
I think it's more correct to separate init from aliases. 
In this #562  I'm introducing the same configuration that was used on Omarchy, which solves this problem.

```shell
if command -v zoxide &> /dev/null; then
  alias cd="zd"
  zd() {
    if [ $# -eq 0 ]; then
      builtin cd ~ && return
    elif [ -d "$1" ]; then
      builtin cd "$1"
    else
      z "$@" && printf "\U000F17A9 " && pwd || echo "Error: Directory not found"
    fi
  }
fi
```

ki11e6:
Thanks for the feedback @Kasui92! Your approach from Omarchy is actually much better. 

Happy to defer to PR #562 if it's addressing this more comprehensively. Should I close this PR in favor of that one?

Kasui92 (commented):
I think so, that might be enough, even just to have a similar alignment with what's being done on Omarchy.

